PR #1, opened 2024-02-19
Login UI

Mooky4591:
Created Login UI and set up the hide/show password functionality. Additional UI changes are needed.


app/src/main/java/com/example/taskyy/data/repository/LoginRepository.kt
```diff
@@ -0,0 +1,4 @@
+package com.example.taskyy.data.repository
+
+class LoginRepository {
+}
```
philipplackner, line 4: I'd bundle multiple related calls in one repository, otherwise you'll end up with so many repositories quickly losing the overview. Something like `AuthRepository` for login, logout and authenticate makes sense :)

app/src/main/java/com/example/taskyy/domain/LoginValidator.kt
```diff
@@ -0,0 +1,4 @@
+package com.example.taskyy.domain
+
+interface LoginValidator {
+}
```
philipplackner, line 4: In my experience, there's not much value in abstracting out business logic. Abstractions only make sense if you have or plan to have at least 2 implementations. But, for business logic I just don't see that happen, even for testing, so this just adds additional boilerplate code to your project.

app/src/main/java/com/example/taskyy/ui/LoginActivity.kt
```diff
@@ -0,0 +1,16 @@
+package com.example.taskyy.ui
+
+import android.os.Bundle
+import androidx.activity.ComponentActivity
+import androidx.activity.compose.setContent
+
+class LoginActivity : ComponentActivity() {
```
philipplackner, line 7: If you're using Compose, you're totally fine with a single activity, so this could just be a generic `MainActivity`

app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
```diff
@@ -0,0 +1,166 @@
...
+@Composable
+fun MainScreen() {
+    Column(
+        horizontalAlignment = Alignment.CenterHorizontally,
+        verticalArrangement = Arrangement.SpaceAround,
+        modifier = Modifier
+            .fillMaxHeight()
+            .background(Color.Black)
+    ) {
+        AddWelcomeBackText()
```
philipplackner, line 46: This should just be `WelcomeBackText`, right?

app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
```diff
@@ -0,0 +1,166 @@
...
+        fontSize = 30.sp,
+        fontWeight = FontWeight.Bold
+    )
+}
+@OptIn(ExperimentalMaterial3Api::class)
+@Composable
+fun CreateUserNameField() {
+    var textFieldState by remember {
+        mutableStateOf("")
+    }
```
philipplackner, line 88: I would always put all UI states in your ViewModel. Using them here with remember will clear the states after a configuration change like a screen rotation

philipplackner (approved):

app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
Mooky4591: I know that using a view model instance in the UI also isn't great for testing so if the state is in the view model, what is the best way to get it to the composable? Is it best to create that view model instance in the activity and then grab the state from the main activity using that instance and pass it as a parameter to the composable? In my case I would pass it to the MainScree() composable?

philipplackner: This is what I suggest:
```
// In your ViewModel
@HiltViewModel
class SampleViewModel @Inject constructor(): ViewModel() {
    private val _state = MutableStateFlow(SampleState()) // Alternatively, Compose state also works
    val state = _state.asStateFlow()

    fun onEvent(event: SampleEvent) {
        when(event) {
            is SampleEvent.Event1 -> // Handle event1
            is SampleEvent.Event2 -> // Handle event2
            // ...
        }
    }

    // ...
}

// In your NavHost
composable(Routes.sample) {
    val viewModel = hiltViewModel<SampleViewModel>()
    val state by viewModel.state.collectAsStateWithLifecycle()
    SampleScreen(
        state = state,
        onEvent = viewModel::onEvent
    )
}

// Your screen composable
@Composable
fun SampleScreen(
    state: SampleState,
    onEvent: (SampleEvent) -> Unit
) {
    // ...
    Button(onClick = { onEvent(SampleEvent.ClickedButtonX) } {
        Text("Click me")
    }
}
```
The screen composable where the real UI is located remains totally testable and previewable, since it doesn't depend on a VM instance :)

Mooky4591: This is my first time working with JetPack Compose and states so pardon the seemingly simple question. You have "SampleState()" being passed as a parameter to the MutableStateFlow(). Where does that come from? I'm trying to implement this but I'm not sure what needs to go there.

Mooky4591: Where to do you recommend the nav host go? Should that be in the MainActivity or should that go with the LoginScreen Composable file?

philipplackner: Separate composable usually, called `NavigationRoot` or similar, I just put it here for simplicity :)

PR #2, opened 2024-02-23
Attempting to pass the state and events back to the viewmodel

Mooky4591:
I am attempting to pass the state and the event to and from the view model but I'm having a hard time figuring out how to do that. I feel like I'm on the right track but need a little guidance

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -64,6 +64,7 @@ dependencies {
     implementation 'androidx.constraintlayout:constraintlayout:2.1.4'
     implementation 'androidx.lifecycle:lifecycle-livedata-ktx:2.7.0'
     implementation 'androidx.lifecycle:lifecycle-viewmodel-ktx:2.7.0'
+    implementation 'androidx.room:room-ktx:2.6.1'
     testImplementation 'junit:junit:4.13.2'
     androidTestImplementation 'androidx.test.ext:junit:1.1.5'
     androidTestImplementation 'androidx.test.espresso:espresso-core:3.5.1'
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,6 +1,9 @@
 plugins {
     id 'com.android.application'
     id 'org.jetbrains.kotlin.android'
+    id 'kotlin-kapt'
+    id 'com.google.dagger.hilt.android'
+    id 'dagger.hilt.android.plugin'
 }
 
 android {
@@ -27,18 +30,18 @@ android {
         }
     }
     compileOptions {
-        sourceCompatibility JavaVersion.VERSION_1_8
-        targetCompatibility JavaVersion.VERSION_1_8
+        sourceCompatibility JavaVersion.VERSION_18
+        targetCompatibility JavaVersion.VERSION_18
     }
     kotlinOptions {
-        jvmTarget = '1.8'
+        jvmTarget = '18'
     }
     buildFeatures {
         compose true
         viewBinding true
     }
     composeOptions {
-        kotlinCompilerExtensionVersion '1.3.2'
+        kotlinCompilerExtensionVersion '1.5.10'
     }
     packagingOptions {
         resources {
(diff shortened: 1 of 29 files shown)


app/build.gradle
```diff
@@ -68,4 +71,29 @@ dependencies {
...
+    implementation "androidx.navigation:navigation-fragment-ktx:$nav_version"
+    implementation "androidx.navigation:navigation-ui-ktx:$nav_version"
+
+    // Jetpack Compose Integration
+    implementation "androidx.navigation:navigation-compose:$nav_version"
+
+    //hilt dependencies
+    implementation 'androidx.hilt:hilt-work:1.1.0'
+    implementation "com.google.dagger:hilt-android:$hilt_version"
+    kapt "com.google.dagger:hilt-compiler:$hilt_version"
```
philipplackner, line 93: I recommend to use KSP (Kotlin Symbol Processing), which is a faster alternative to kapt. Both Dagger and Room support that.

I explain it here: https://www.youtube.com/watch?v=WVhgUGyUnhQ

app/src/main/java/com/example/taskyy/app.java
```diff
@@ -0,0 +1,9 @@
+package com.example.taskyy;
+
+import android.app.Application;
+
+import dagger.hilt.android.HiltAndroidApp;
+
+@HiltAndroidApp
+public class app extends Application {
```
philipplackner: Class names should be capitalized as per convention :)

app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
```diff
@@ -33,17 +33,19 @@ import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.unit.dp
 import androidx.compose.ui.unit.sp
 import com.example.taskyy.R
+import com.example.taskyy.ui.events.LoginEvents
+import kotlin.reflect.KFunction1
 
 @Composable
-fun MainScreen() {
+fun MainScreen(state: String, onEvent: KFunction1<LoginEvents, Unit>) {
```
philipplackner: The state shouldn't be a string here, you'd just pass the normal instance and a normal Kotlin lambda:
```
fun MainScreen(state: LoginState, onEvent: (LoginEvent) -> Unit)
```

app/src/main/java/com/example/taskyy/ui/events/LoginEvents.kt
```diff
@@ -0,0 +1,9 @@
+package com.example.taskyy.ui.events
+
+class LoginEvents {
+    companion object {
+        fun textChanged(): LoginEvents {
+            return LoginEvents()
+        }
+    }
+}
```
philipplackner: This class doesn't make much sense to me, usually you implement the UI-to-ViewModel events like this:
```
sealed interface LoginEvent {
    data class OnEmailChanged(val email: String): LoginEvent
    data object OnLoginClick: LoginEvent
    ...
}
```
And in your ViewModel:
```
fun onEvent(event: LoginEvent) {
    when(event) {
        is LoginEvent.OnEmailChanged -> state = state.copy(email = event.email)
        is LoginEvent.Login -> login()
        ...
    }
}
```

philipplackner (approved):

app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
Mooky4591: So I need to create a LoginState class? Then use that as the type when creating the state in the view model?

app/src/main/java/com/example/taskyy/ui/events/LoginEvents.kt
Mooky4591: Why use 'data' here? Also why use object for the click but class for the email change? Is it because you're updating the state value?

app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
philipplackner: Correct, something like this:
```
data class LoginState(
    val email: String = "",
    val password: String = "",
    val isLoggingIn: Boolean = false,
    ...
)
```
and in the VM:
```
var state by mutableStateOf(LoginState())
    private set 
```
Alternatively, StateFlow is also fine

app/src/main/java/com/example/taskyy/ui/events/LoginEvents.kt
philipplackner: Data classes and objects just come with already implemented `equals`, `hashCode()` and `copy()` functions which is very convenient.
Normally, we'd expect that these are the same:
```
val class1 = MyClass("HelloWorld")
val class2 = MyClass("HelloWorld")
println(class1 == class")
```
But this will only print true if `MyClass` is a data class, since it will then compare their values rather than whether they're the same reference (they're two different instances, so with a normal class the comparison would return false).

And whenever a class doesn't need any constructor arguments and doesn't have any internal state, you can use an object instead

PR #3, opened 2024-02-24
Created the registration classes and started to set up the registrati…

Mooky4591:
…on screen. Is there a way to accept any type of state in a composable? For example, I have a composable created the email field. It accepts a LoginState. I need to create a email field for the registration screen. Is there a way to reuse that CreateEmailField or do I need to create another composable that does the same thing but accepts a registartionstate?


app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
```diff
@@ -0,0 +1,54 @@
...
+            route = "login_screen"
+        ) {
+            composable(route = Screen.Login.route) {
+                val viewModel = hiltViewModel<LoginViewModel>()
+                val state = viewModel.state
+                LoginScreen(state, viewModel::onEvent, navController)
+            }
+            composable(route = Screen.Register.route) { entry ->
+                val loginViewModel = entry.loginViewModel<LoginViewModel>(navController = navController)
+                val state = loginViewModel.state
```
philipplackner: I assume this is temporary since you've asked a question about this. In case not, you'd rather want to give the RegisterScreen its own ViewModel to keep the responsibilities separate. Login doesn't need to know anything about registration and registration doesn't need to know anything about login, yet with one shared VM you mix these together.

Shared ViewModels only make sense if you need to share state between multiple screens. Here each screen has its own state, so that's not needed

app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
```diff
@@ -0,0 +1,54 @@
...
+@Composable
+inline fun <reified T : ViewModel> NavBackStackEntry.loginViewModel(
+    navController: NavController,
+): T {
+    val navGraphRoute = destination.parent?.route ?: return hiltViewModel()
+    val parentEntry = remember(this) {
+        navController.getBackStackEntry(navGraphRoute)
+    }
+    return hiltViewModel(parentEntry)
+}
```
philipplackner, line 54: You made this function use Generics, but then called it `loginViewModel`. In the end, you could need this for more than just the login VM, so I'd prefer a naming that communicates what this function actually does. So in this case, something like `sharedViewModel` would be a better pick

app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
```diff
@@ -56,49 +60,57 @@ fun LoginScreen(state: LoginState, onEvent: (LoginEvent) -> Unit) {
...
+        text = text,
         color = Color.White,
         fontSize = 30.sp,
         fontWeight = FontWeight.Bold
     )
 }
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
-fun CreateUserNameField(state: LoginState, onEvent: (LoginEvent) -> Unit) {
+fun CreateEmailField(state: LoginState, onEvent: (LoginEvent) -> Unit) {
```
philipplackner: To answer your question:
You wouldn't want to make this field dependent on `LoginState` since as you realized this won't make it reusable for screens where you don't have a `LoginState`. Instead, only give this screen the fields it needs. For a text field, that might be the text itself, whether the field is valid and maybe a hint.
That way, you can also call this something more generic, like `TaskyTextField` which just styles it based on the mockups, but allows you to use it for different types of text fields

app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
```diff
@@ -158,15 +173,21 @@ fun CreateHidePasswordToggle(state: LoginState, onEvent: (LoginEvent) -> Unit) {
 }
 
 @Composable
-fun CreateLoginButton() {
+fun CreateLoginButton(state: LoginState, onEvent: (LoginEvent) -> Unit, navController: NavController, text: String) {
```
philipplackner: Same here, I'd only pass down to the button what it really needs. It doesn't need to know the whole login screen's UI, but rather just a text and a lambda when it's clicked. Then you can again reuse it for different purposes

philipplackner (approved):

app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
Mooky4591: If I were to not include that state here, how would I update the onValueChange? I use that to update the email in the view model.

philipplackner: Not 100% sure what you mean, but you'd just pass an `onValueChange: (String) -> Unit` lambda to this text field. In your screen, you can then do this:
```
CreateEmailField(
    text = state.text,
    onValueChange = { onEvent(LoginEvent.EmailTextChanged(it) },
    ...
)
```

Mooky4591: In your example your using state.text to set the text value. But you don't have state so when the user updates the email field, if I don't have the state, how do I update the textfield?

app/src/main/java/com/example/taskyy/domain/repository/AuthRepository.kt
Mooky4591, line 1: I'm doing email and name validation in the AuthRepository, is that correct? I wasn't sure if this should go here or in a use case

app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
Mooky4591: Ok I see how to not pass the state but what about specific events? It accepts a LoginEvent to the LoginEvent.OnEmailChanged(it) method.

philipplackner: For each event, you simply pass a lambda just like when the text changed. For example, this could also be done:
```
CreateEmailField(
    text: String,
    onValueChange: (String) -> Unit,
    onEndIconClick: () -> Unit
)
```
Unless you have a component that really needs most of the events declared in the screen event class (here `LoginEvent`), I'd prefer using separate lambdas to make this composable independent of the screen and reusable therefore

app/src/main/java/com/example/taskyy/domain/repository/AuthRepository.kt
philipplackner: The job of a repository is to implement one or multiple data sources, possibly combine them and decide which data stream to return. Validation is just a piece of business logic, so that should rather go in a separate class like a use case

app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
Mooky4591: I'm sorry but I'm not understanding this. If I don't pass an event into the composable, how would onEvent in the view model know what to do

philipplackner: Your composable will look like this:
```
@Composable
CreateEmailField(
    text: String,
    onValueChange: (String) -> Unit,
)
```
Then, you add this to your `LoginScreen` for example (which has the `onEvent` lambda):
```
// ... Content before
CreateEmailField(
    text = state.text,
    onValueChange = { email ->
        onEvent(LoginEvent.OnEmailChange(email)
    }
)
// ... Content after
```
The onEvent lambda still exists, you just don't add it to the CreateEmailField composable, since all you care about is just one or two evens (like when the text changes).
Now you can use the text field wherever you want, since it's not dependent on `LoginEvent` specifically anymore

Mooky4591: When I try to use the onEvent lambda, it doesn't recognize it if I don't pass it into the composable. It says that onEvent is an unresolved reference. It also says I'm not using the onValueChange parameter that I'm passing in. I must be doing something incorrectly

philipplackner: Can you post the code as you have it right now? The whole screen Composable as well as the whole text field composable

PR #4, opened 2024-02-26
moved email validation to use case. attempting to make composable fun…

Mooky4591:
…ctions reusable

### DIFF
--- a/app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
+++ b/app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
@@ -61,7 +61,7 @@ fun LoginScreen(state: LoginState, onEvent: (LoginEvent) -> Unit, navController:
                     .fillMaxHeight(.75f)
                     .padding(30.dp)
             ) {
-                CreateEmailField(state.isEmailValid, onEvent, state.email)
+                CreateEmailField(state.isEmailValid, (String) -> Unit,  state.email)
                 Spacer(modifier = Modifier.height(10.dp))
                 CreatePasswordField(state.password, onEvent, state.isPasswordVisible)
                 Spacer(modifier = Modifier.height(10.dp))
@@ -87,11 +87,14 @@ fun TopText(text: String) {
 }
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
-fun CreateEmailField(isEmailValid: Boolean, onEvent: (LoginEvent) -> Unit, email: String) {
+fun CreateEmailField(isEmailValid: Boolean, onValueChange: (String) -> Unit, email: String) {
     TextField(
         value = email,
         onValueChange = {
-            onEvent(LoginEvent.OnEmailChanged(it))
+            onValueChange(
+                email ->
+            )
+
         },
         placeholder = {
             Text(text = stringResource(R.string.email_address), color = Color.LightGray)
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
```diff
@@ -87,11 +87,14 @@ fun TopText(text: String) {
...
 @Composable
-fun CreateEmailField(isEmailValid: Boolean, onEvent: (LoginEvent) -> Unit, email: String) {
+fun CreateEmailField(isEmailValid: Boolean, onValueChange: (String) -> Unit, email: String) {
     TextField(
         value = email,
         onValueChange = {
-            onEvent(LoginEvent.OnEmailChanged(it))
+            onValueChange(
+                email ->
+            )
```
philipplackner, line 96: Here the syntax also isn't correct. You call a lambda like a normal function, in this case: `onValueChange(it)`

app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
```diff
@@ -61,7 +61,7 @@ fun LoginScreen(state: LoginState, onEvent: (LoginEvent) -> Unit, navController:
                     .fillMaxHeight(.75f)
                     .padding(30.dp)
             ) {
-                CreateEmailField(state.isEmailValid, onEvent, state.email)
+                CreateEmailField(state.isEmailValid, (String) -> Unit,  state.email)
```
philipplackner, line 64: Here, the syntax is not correct. `(String) -> Unit` is correct when defining a lambda inside the parameter list, but when creating a lambda reference, you can just to this with curly braces. In this example, it would need to be like this:
```
CreateEmailField(
    state.isEmailValid,
    { email -> onEvent(LoginEvent.OnEmailChange(email) },
    state.email
)
```

app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
```diff
@@ -116,12 +119,12 @@ fun CreateEmailField(isEmailValid: Boolean, onEvent: (LoginEvent) -> Unit, email
 
 @OptIn(ExperimentalMaterial3Api::class)
 @Composable
-fun CreatePasswordField(password: String, onEvent: (LoginEvent) -> Unit, isPasswordVisible: Boolean) {
+fun CreatePasswordField(password: String, onEvent: () -> Unit, isPasswordVisible: Boolean) {
```
philipplackner, line 122: Just like for the email field, you'd also just provide a `onValueChange: (String) -> Unit` lambda here

app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
```diff
@@ -146,7 +149,7 @@ fun showOrHidePassword(isPasswordVisible: Boolean): VisualTransformation {
 
 
 @Composable
-fun CreateHidePasswordToggle(isPasswordVisible: Boolean, onEvent: (LoginEvent) -> Unit) {
+fun CreateHidePasswordToggle(isPasswordVisible: Boolean, onEvent: () -> Unit) {
```
philipplackner, line 152: Here, the lambda would need to look like this: `onTogglePassword: () -> Unit`

philipplackner (approved):

PR #5, opened 2024-02-26
Attempted to make the composable functions interchangeable

Mooky4591:
I was able to figure out the text fields updating correctly and making them reusable. I do have questions about the login button though. How do I pass the click back?

### DIFF
--- a/app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
+++ b/app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
@@ -36,7 +36,6 @@ import androidx.navigation.NavController
 import com.example.taskyy.R
 import com.example.taskyy.domain.navigation.Screen
 import com.example.taskyy.ui.events.LoginEvent
-import com.example.taskyy.ui.events.RegisterEvent
 import com.example.taskyy.ui.viewmodels.LoginState
 
 @Composable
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
```diff
@@ -175,13 +174,13 @@ fun CreateHidePasswordToggle(isPasswordVisible: Boolean, onEvent: () -> Unit) {
...
     Button(
         onClick = {
-                  if(onLoginEvent != null) {
-                    //onLoginEvent(LoginEvent.OnLoginClick)
+                  if(text == "LOG IN") {
+                      onClick()
                   } else {
-                     //onRegisterEvent(RegisterEvent.OnGetStartedClick)
+                     onClick()
                   }
```
philipplackner, line 184: Few things:
1. First, I'd rename the button to something like `TaskyActionButton`, so it's more generic and shows that it's intended to be used for more than just log in.
2. In your `onClick` lambda, all you need to call is the `onClick()` lambda from your parameters. No need for any if-checks, since the screen that adds this button already knows what it is for. Just the button doesn't need to know that.

philipplackner (approved):

app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
```diff
@@ -161,11 +162,9 @@ fun CreateHidePasswordToggle(isPasswordVisible: Boolean, onEvent: () -> Unit) {
 
     // Toggle button to hide or display password
     IconButton(onClick = {
-        onEvent(
             if (isPasswordVisible) {
-                LoginEvent.OnTogglePasswordVisibility(false)
-            } else LoginEvent.OnTogglePasswordVisibility(true)
-        )
+                onClick(false)
+            } else onClick(true)
```
philipplackner, line 167: Could be shortened to `onClick(!isPasswordVisible)`

philipplackner (approved):

PR #6, opened 2024-02-27
2 26 24 creating room database

Mooky4591:
Started to connect the database to the viewmodel to start creating user objects for registering users. I am having an issue getting the context that is needed for creating the userDao. Also, should I abstract the DB further?


app/src/main/java/com/example/taskyy/MainActivity.kt
```diff
@@ -8,10 +9,11 @@ import dagger.hilt.android.AndroidEntryPoint
 
 @AndroidEntryPoint
 class MainActivity : ComponentActivity() {
+    val context: Context = this
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContent {
-            Nav()
+            Nav(context)
```
philipplackner: This is not necessary, you can get the context in any Composable function with `val context = LocalContext.current`

app/src/main/java/com/example/taskyy/data/room_database/TaskyyDatabase.kt
```diff
@@ -0,0 +1,35 @@
...
+
+    companion object {
+
+        @Volatile
+        private var INSTANCE: RoomDatabase? = null
+        fun getDatabase(context: Context): RoomDatabase{
+            val tempInstance = INSTANCE
+            if(tempInstance != null) {
+                return tempInstance
+            }
```
philipplackner: I think you could shorten this to
```
if(INSTANCE != null) {
    return INSTANCE
}
```

app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
```diff
@@ -36,14 +37,14 @@ fun Nav() {
...
             }
         }
     }
 }
 @Composable
-inline fun <reified T : ViewModel> NavBackStackEntry.loginViewModel(
+inline fun <reified T : ViewModel> NavBackStackEntry.sharedViewModel(
     navController: NavController,
 ): T {
     val navGraphRoute = destination.parent?.route ?: return hiltViewModel()
```
philipplackner: A shared VM only makes sense if you need to share state between multiple screens. However, for authentication that's not really needed, since each screen has its own state. Sharing the state will technically work, but it mixes responsibilities together in one ViewModel. 
Instead, I'd recommend to have a separate `LoginViewModel` and `RegisterViewModel` which are only bound to their corresponding screen.

app/src/main/java/com/example/taskyy/ui/events/RegisterEvent.kt
```diff
@@ -1,12 +1,14 @@
...
+
 
 sealed interface RegisterEvent{
     data class OnNameChanged(val name: String): RegisterEvent
     data class OnEmailChanged(val email: String): RegisterEvent
     data class OnPasswordChanged(val password: String): RegisterEvent
     data class OnTogglePasswordVisibility(val isPasswordVisible: Boolean): RegisterEvent
 
-    data object OnGetStartedClick: RegisterEvent
+    data class OnGetStartedClick(val applicationContext: Context) : RegisterEvent
```
philipplackner: I wouldn't recommend to pass the context to a ViewModel. On the one hand, it can quickly happen that you run into memory leaks in the VM when saving it in a field there (not with the app context, but with the activity context).
But even with the application context, you have the issue that your ViewModel suddenly relies on the Android SDK. In case you'd want to write a test case for it later, that would only work when it's an instrumented test (which is a test that has to run on an actual Android device and is much slower) rather than a local unit test (which is a test that simply runs on the JVM directly and doesn't depend on Android)

app/src/main/java/com/example/taskyy/ui/viewmodels/RegisterViewModel.kt
```diff
@@ -21,21 +24,24 @@ class RegisterViewModel @Inject constructor(
...
             is RegisterEvent.OnTogglePasswordVisibility -> state = state.copy(isPasswordVisible = event.isPasswordVisible)
-                   }
+        }
     }
 
-    private fun register() {
-        authRepository.registerUser(name = state.name, password = state.password, email = state.email)
+    private fun register(context: Context) {
+        val user = User(name = state.name, password = state.password, email = state.email)
+        val dao = TaskyyDatabase.getDatabase(context = context).userDao()
```
philipplackner: Here I'd recommend to get into proper dependency injection, so you would pass a reference to this DB (or better just the DAO you need here) in the constructor of the ViewModel, so it remains flexible what you pass for that.
A good starting point to get into that is my course on Dagger-Hilt:
https://www.youtube.com/watch?v=bbMsuI2p1DQ

app/src/main/java/com/example/taskyy/ui/viewmodels/RegisterViewModel.kt
```diff
@@ -21,21 +24,24 @@ class RegisterViewModel @Inject constructor(
...
 
-    data class RegisterState(
+data class RegisterState(
         var email: String = "",
         var password: String = "",
         var name: String = "",
         var isEmailValid: Boolean = false,
-        var isPasswordVisible: Boolean = false
+        var isPasswordVisible: Boolean = false,
+        val context: Context? = null
```
philipplackner: If this was the activity context, you'd cause a memory leak here. The whole point of the VM is to outlive the lifecycle of the activity, but when keeping a reference to the activity in the VM, the garbage collector won't be able to free up this reference when the activity is destroyed (like on a screen rotation).
So you would effectively keep a reference to an activity that isn't used anymore. I know you only passed the app context here, but this is very dangerous

philipplackner (approved):

PR #7, opened 2024-02-28
Removed shared viewmodel function, removed passing the context to the…

Mooky4591:
… viewmodel. Attempting to pass the UserDao through dependency injection but having a hard time accessing the context in the AuthModule

### DIFF
--- a/app/src/main/java/com/example/taskyy/data/room_database/TaskyyDatabase.kt
+++ b/app/src/main/java/com/example/taskyy/data/room_database/TaskyyDatabase.kt
@@ -16,14 +16,14 @@ abstract class TaskyyDatabase: RoomDatabase() {
 
         @Volatile
         private var INSTANCE: RoomDatabase? = null
-        fun getDatabase(context: Context): RoomDatabase{
+        fun getDatabase(context: Context): TaskyyDatabase{
             if(INSTANCE != null) {
-                return INSTANCE as RoomDatabase
+                return INSTANCE as TaskyyDatabase
             }
             synchronized(this){
                 val instance = Room.databaseBuilder(
                     context.applicationContext,
-                    RoomDatabase::class.java,
+                    TaskyyDatabase::class.java,
                     "database"
                 ).build()
                 INSTANCE = instance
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/example/taskyy/data/room_database/TaskyyDatabase.kt
```diff
@@ -16,14 +16,14 @@ abstract class TaskyyDatabase: RoomDatabase() {
 
         @Volatile
         private var INSTANCE: RoomDatabase? = null
-        fun getDatabase(context: Context): RoomDatabase{
+        fun getDatabase(context: Context): TaskyyDatabase{
             if(INSTANCE != null) {
-                return INSTANCE as RoomDatabase
+                return INSTANCE as TaskyyDatabase
             }
```
philipplackner, line 22: If you change the type for INSTANCE to `TaskyyDatabase`, you don't need that type cast here :)

app/src/main/java/com/example/taskyy/data/room_entity/UserEntity.kt
```diff
@@ -1,14 +1,15 @@
...
 data class UserEntity(
     @PrimaryKey(autoGenerate = true)
-    val id: Int,
-    val firstName: String,
-    val lastName: String,
-    val phoneNumber: String
+    @ColumnInfo val id: Int,
+    @ColumnInfo val firstName: String,
+    @ColumnInfo val lastName: String,
+    @ColumnInfo val phoneNumber: String
```
philipplackner, line 13: These annotations aren't necessary, by default Room will use the variable name as column name

app/src/main/java/com/example/taskyy/di/AuthModule.kt
```diff
@@ -33,6 +33,7 @@ object AuthModule {
     @Provides
     @Singleton
     fun provideUserDao(): UserDao {
+        //this doesn't work since it needs the composable annotation. How to get the context here?
         return TaskyyDatabase.getDatabase(LocalContext.current).userDao()
     }
```
philipplackner, line 38: You're not in a composable function, which is why this doesn't work. With Dagger, you can do this:
```
@Provides
@Singleton
fun provideUserDao(@ApplicationContext context: Context): UserDao {
    return TaskyyDatabase.getDatabase(context).userDao()
}
```

philipplackner (approved):

PR #8, opened 2024-02-29
Finished creating DB and set up insertion of the newly registerd user…

Mooky4591:
…Entity. Deleted unneeded user object since I am using UserEntity.


app/build.gradle
```diff
@@ -95,6 +96,11 @@ dependencies {
     kapt "com.google.dagger:hilt-compiler:$hilt_version"
     implementation "androidx.hilt:hilt-navigation-compose:1.1.0"
 
+    //room dependencies
+    kapt "androidx.room:room-compiler:$room_version"
```
philipplackner, line 100: I recommend to use KSP (Kotlin Symbol Processing), which is a faster alternative to kapt. Both Dagger and Room support that.

I explain it here: https://www.youtube.com/watch?v=WVhgUGyUnhQ

philipplackner (approved):

PR #9, opened 2024-02-29
02 29 23 set up retrofit

Mooky4591:
Started getting the entities set up to use for parsing the download and as mappers for the room db. Started setting up retrofit. I am having an issue getting the context into the AuthRepositoryImpl. I'm not sure I can inject context there or what I need to do


app/src/main/java/com/example/taskyy/data/local/room_entity/ReminderEntity.kt
```diff
@@ -0,0 +1,12 @@
+package com.example.taskyy.data.local.room_entity
+
+import androidx.room.Entity
+
+@Entity
+data class ReminderEntity(
+    val description: String?,
+    val id: String,
```
philipplackner, line 9: You're missing the primary key annotation here

app/src/main/java/com/example/taskyy/data/local/room_entity/TaskEntity.kt
```diff
@@ -0,0 +1,13 @@
+package com.example.taskyy.data.local.room_entity
+
+import androidx.room.Entity
+
+@Entity(tableName = "entity_table")
+data class TaskEntity(
+    val description: String,
+    val id: String,
```
philipplackner, line 9: Same here

app/src/main/java/com/example/taskyy/data/remote/RetrofitInstance.kt
```diff
@@ -0,0 +1,15 @@
...
+
+object RetrofitInstance {
+
+val api: TaskyyApi by lazy {
+    Retrofit.Builder()
+        .baseUrl("https://tasky.pl-coding.com/")
+        .addConverterFactory(GsonConverterFactory.create())
+        .build()
+        .create(TaskyyApi::class.java)
+    }
```
philipplackner: If you're using Dagger, you would want to provide this instance in your `AuthModule`. Then you can inject it wherever you want

app/src/main/java/com/example/taskyy/data/local/room_database/TaskyyDatabase.kt
```diff
@@ -16,7 +16,7 @@ abstract class TaskyyDatabase: RoomDatabase() {
 
         @Volatile
         private var INSTANCE: TaskyyDatabase? = null
-        fun getDatabase(context: Context): TaskyyDatabase{
+        fun getDatabase(context: Context): TaskyyDatabase {
             if(INSTANCE != null) {
                 return INSTANCE as TaskyyDatabase
             }
```
philipplackner, line 22: With Dagger, you can actually get rid of all that. If you provide this DB instance in your Dagger module, it will take care of managing it as a singleton behind the scenes

app/src/main/java/com/example/taskyy/data/remote/TaskyyApi.kt
```diff
@@ -0,0 +1,77 @@
...
+import retrofit2.http.GET
+import retrofit2.http.Multipart
+import retrofit2.http.POST
+import retrofit2.http.PUT
+import retrofit2.http.Query
+
+interface TaskyyApi {
+
+    @POST("/register")
+    fun registerUser(@Query("key") key: String)
```
philipplackner, line 16: The API key will be needed for every single request. In that case it makes sense to add an interceptor. That's something that lets you manipulate every request before it's fired (for example to add a header like here). This is how it can be done:
```
class ApiKeyInterceptor(
    private val apiKey: String
): Interceptor {

    override fun intercept(chain: Interceptor.Chain): Response {
        return chain.proceed(
            chain
                .request()
                .newBuilder()
                .addHeader("x-api-key", apiKey)
                .build()
        )
    }
}
```
And then add it when creating the API instance:
```
@Provides
@Singleton
fun provideOkHttpClient(): OkHttpClient {
    return OkHttpClient.Builder()
        .addInterceptor(ApiKeyInterceptor(BuildConfig.API_KEY))
        .build()
}

@Provides
@Singleton
fun provideAuthApi(client: OkHttpClient): AuthApi {
    return Retrofit.Builder()
        .baseUrl(BASE_URL)
        .addConverterFactory(MoshiConverterFactory.create())
        .client(client) // <-- Here, you add the client with the interceptor
        .build()
        .create()
}
```

app/src/main/java/com/example/taskyy/data/repositories/AuthRepositoryImpl.kt
```diff
@@ -0,0 +1,30 @@
...
+            userDao.insertUser(userEntity)
+             }
+        }
+    }
+
+    override fun registerUser(userEntity: UserEntity) {
+        runBlocking { launch {
+            val response = try {
+                //how to get context here
+                RetrofitInstance.api.registerUser()
```
philipplackner: Why do you need the context here? Generally there's no issue with injecting the app context into a repository, but I don't think this is what you want here. With Dagger, you could do this:
```
class AuthRepositoryImpl @Inject constructor(private val app: Application)
```
`app` in this case is nothing else than an application context reference. But again, I don't think you need this here

app/src/main/java/com/example/taskyy/domain/repository/AuthRepository.kt
```diff
@@ -1,8 +1,9 @@
...
-import com.example.taskyy.data.data_access_objects.UserDao
-import com.example.taskyy.data.room_entity.UserEntity
+import com.example.taskyy.data.local.data_access_objects.UserDao
+import com.example.taskyy.data.local.room_entity.UserEntity
 
 interface AuthRepository {
-    fun registerUser(userEntity: UserEntity, userDao: UserDao)
+    fun addUserToDatabase(userEntity: UserEntity, userDao: UserDao)
+    fun registerUser(userEntity: UserEntity)
 }
```
philipplackner: Here, you're breaking the principle that the domain layer should be isolated. Isolation here means that domain shouldn't know about any implementation details (**how** something is implemented, e.g. which library is used for HTTP calls, how data is stored in a DB or how your request/response classes look like).
In this interface which sits in domain (that itself is correct), you're accessing classes from data however which is not allowed. The idea behind this isolation is that frequently changing code (implementation details) doesn't cause changes in less frequently changing code (business logic). If you break this principle like here, for example if you'd change your `UserEntity`, this would directly cause a change here in the repo. 

Instead, only pass data to the repository that's needed there. For the `registerUser` function for example, all the repository needs to know is the full name, email and password the user wants to register with. The DAO you can just inject in the repository then:
```
class AuthRepositoryImpl(
    private val userDao: UserDao,
    private val taskyApi: TaskyyApi
): AuthRepository {
    // ...
}
```
And provide the repository like this in your AppModule:
```
@Provides
@Singleton
fun provideAuthRepository(userDao: UserDao, api: TaskyyApi): AuthRepository {
    return AuthRepositoryImpl(userDao, api)
}
```
Now that Dagger knows how to create such a repository reference, you can inject this repository wherever you want (typically in a ViewModel):
```
@HiltViewModel
class RegisterViewModel @Inject constructor(
    private val repository: AuthRepository
): ViewModel() { ... }
```
I hope this long comment wasn't too confusing. If it was, please ask 😄

philipplackner (approved):

app/src/main/java/com/example/taskyy/data/repositories/AuthRepositoryImpl.kt
```diff
@@ -0,0 +1,36 @@
...
+            userDao.insertUser(userEntity)
+             }
+        }
+    }
+
+    override fun registerUser(userEntity: UserEntity) {
+        runBlocking { launch {
+            val response = try {
+                //how to use the provideOkHttp and provideAuthApi here?
+                taskyyApi.registerUser()
```
Mooky4591, line 29: I'm not totally sure how to use the interceptor here to provide the key. Since the API Key is a query parameter, it is required here.

Mooky4591: I think I'm getting confused between using the TaskyyApi object and the RetrofitInstance. I'm confused at what each one does now

PR #10, opened 2024-03-01
02 29 23 set up retrofit

Mooky4591:
I've messed up the register API call somewhere. I'm not sure where, I am getting an HTTPException. I am assuming I did something wrong with the interceptor.

### DIFF
--- a/app/src/main/java/com/example/taskyy/data/local/room_entity/UserEntity.kt
+++ b/app/src/main/java/com/example/taskyy/data/local/room_entity/UserEntity.kt
@@ -6,9 +6,8 @@ import androidx.room.PrimaryKey
 @Entity(tableName = "user_table")
 data class UserEntity(
     @PrimaryKey(autoGenerate = true)
-    val id: Int? = null,
-    val name: String,
-    val password: String,
-    val email: String
+    var id: Int? = null,
+    var name: String,
+    var email: String
 ) {
 }
--- a/app/src/main/java/com/example/taskyy/data/remote/TaskyyApi.kt
+++ b/app/src/main/java/com/example/taskyy/data/remote/TaskyyApi.kt
@@ -4,6 +4,8 @@ import com.example.taskyy.data.local.room_entity.EventEntity
 import com.example.taskyy.data.local.room_entity.ReminderEntity
 import com.example.taskyy.data.local.room_entity.TaskEntity
 import retrofit2.http.DELETE
+import retrofit2.http.Field
+import retrofit2.http.FormUrlEncoded
 import retrofit2.http.GET
 import retrofit2.http.Multipart
 import retrofit2.http.POST
@@ -11,9 +13,13 @@ import retrofit2.http.PUT
 import retrofit2.http.Query
 
 interface TaskyyApi {
-
+    @FormUrlEncoded
     @POST("/register")
-    suspend fun registerUser()
+    suspend fun registerUser(
+        @Field("fullName") fullName: String,
+        @Field("email") email: String,
+        @Field("password") password: String
+    )
 
     @POST("/login")
     fun loginUser(): retrofit2.Response<LoginUserResponse>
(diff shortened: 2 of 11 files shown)


app/src/main/java/com/example/taskyy/data/remote/TaskyyApi.kt
```diff
@@ -4,19 +4,25 @@ import com.example.taskyy.data.local.room_entity.EventEntity
...
 
 interface TaskyyApi {
-
+    @FormUrlEncoded
     @POST("/register")
-    fun registerUser(@Query("key") key: String)
+    suspend fun registerUser(
+        @Field("fullName") fullName: String,
+        @Field("email") email: String,
+        @Field("password") password: String
```
philipplackner: The /register endpoint expects the request as a JSON body. You're attaching the strings as separate fields here. To fix that you can do this:
```
// In auth->data->request for example
data class RegisterRequest(
    val fullName: String,
    val email: String,
    val password: String
)
```
And then your Retrofit function:
```
@POST("/register")
suspend fun registerUser(@Body body: RegisterRequest)
```

app/src/main/java/com/example/taskyy/data/repositories/AuthRepositoryImpl.kt
```diff
@@ -25,9 +25,11 @@ class AuthRepositoryImpl @Inject constructor(
     override fun registerUser(userEntity: UserEntity) {
```
philipplackner: I've commented on that in detail in the last PR, but here you're still coupling your domain layer to your data layer which isn't allowed. `AuthRepository` which is in domain knows about `UserEntity` which is in data. Only the other way of communication is allowed, that data would know something from domain. 
To fix that, you could create a class like this in domain, e.g.:
```
data class Registration(
    val fullName: String,
    val email: String,
    val password: String
)
```
and then pass that to this `registerUser` function instead of the entity. The entity doesn't feel like the right type here anyways as it contains the password which you would rather not want to save in a local database

philipplackner (approved):

app/src/main/java/com/example/taskyy/domain/objects/RegisterUser.kt
```diff
@@ -0,0 +1,7 @@
+package com.example.taskyy.domain.objects
+
+data class RegisterUser(
+    val fullName: String,
+    val email: String,
+    val password: String
+) {}
```
philipplackner, line 7: Strictly speaking, you should also have a data model of this class. While this will work with the JSON parsing itself here, a change in the JSON field names would require a change in this class (and therefore your domain layer) which isn't allowed. 
In your mapper, you map it to a `UserEntity`, however "Entity" is usually what we call objects in a local database and not those we use for HTTP communication. Those are actually called DTO (Data Transfer Object). So what your mapper should do is to map this `RegisterUser` domain model to something like a `RegisterUserDto` data model which in this case looks the same as this class (coincidence, that's not always the case - domain and data models very often look differently). Then you make the request with that data model

app/src/main/java/com/example/taskyy/data/local/room_entity/UserEntity.kt
```diff
@@ -6,9 +6,8 @@ import androidx.room.PrimaryKey
 @Entity(tableName = "user_table")
 data class UserEntity(
     @PrimaryKey(autoGenerate = true)
-    val id: Int? = null,
-    val name: String,
-    val password: String,
-    val email: String
+    var id: Int? = null,
+    var name: String,
+    var email: String
```
philipplackner, line 11: This is likely why you're not getting any response from the server. The class you send to the server needs to match exactly what the server expects. Same fields and same field names. When looking at the API docs, the /register endpoint accepts this format 
```
{
    "fullName": <String>
    "email": <String>
    "password": <String>
}
```
This doesn't match with your `UserEntity` which you're currently sending to the server. Check my next comment on how to fix that

philipplackner (approved):

PR #11, opened 2024-03-05
changed the registerUser class to RegisterUserDTO. I stopped converti…

Mooky4591:
…ng the DTO to an Entity and am using the DTO to send the HTTP request. I'm fairly certain the DTO matches the fields the endpoint is looking for.

### DIFF
--- a/app/src/main/java/com/example/taskyy/domain/data_transfer_objects/RegisterUserDTO.kt
+++ b/app/src/main/java/com/example/taskyy/domain/data_transfer_objects/RegisterUserDTO.kt
@@ -1,0 +1,7 @@
+package com.example.taskyy.domain.data_transfer_objects
+
+data class RegisterUserDTO(
+    val fullName: String,
+    val email: String,
+    val password: String
+) {}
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/example/taskyy/domain/data_transfer_objects/RegisterUserDTO.kt
```diff
@@ -0,0 +1,7 @@
+package com.example.taskyy.domain.data_transfer_objects
```
philipplackner, line 1: As this is a data model, this should also go in your data package. This is something domain should not know about :)

app/src/main/java/com/example/taskyy/domain/usecases/RegisterUseCase.kt
```diff
@@ -1,14 +1,14 @@
...
 
 class RegisterUseCase @Inject constructor(
     private val authRepository: AuthRepository
 ) {
 
-    fun registerUser(registerUser: RegisterUser) {
-        authRepository.registerUser(registerUser)
+    fun registerUser(registerUserDTO: RegisterUserDTO) {
+        authRepository.registerUser(registerUserDTO)
     }
```
philipplackner, line 13: This was correct as it was before. In the last PR I mentioned that you'll need a domain model which you can use across the app and a data model (the DTO) you can only use in data which contains the JSON specification. This use case should only know the domain model therefore

philipplackner (approved):

PR #12, opened 2024-03-08
Created a user object that can be used in the domain layer. Then mapp…

Mooky4591:
…ed it to a DTO for API usage. I am getting an HTTP exception


app/src/main/java/com/example/taskyy/data/mappers/UserToUserDTOMapper.kt
```diff
@@ -0,0 +1,13 @@
...
+import com.example.taskyy.domain.objects.User
+import javax.inject.Inject
+
+class UserToUserDTOMapper @Inject constructor(
+) {
+
+    fun mapUserToUserDTO(user: User): RegisterUserDTO {
+        return RegisterUserDTO(fullName = user.fullName, email = user.email, password = user.password)
+    }
+}
```
philipplackner, line 13: I personally never found use in making mappers a separate class. I love just making them normal extension functions: 
```
fun User.toRegisterUserDTO(): RegisterUserDTO {
    // ...
}
```

app/src/main/java/com/example/taskyy/data/repositories/AuthRepositoryImpl.kt
```diff
@@ -23,14 +24,20 @@ class AuthRepositoryImpl @Inject constructor(
...
                 Log.e("TAG", "HttpException, " + e.printStackTrace())
             }
             Log.e("TAG", response.toString())
         } }
     }
+
+    override fun validatePassword(password: String): Boolean {
+        val passwordRegex = "^(?=.*[A-Za-z])(?=.*\\d)[A-Za-z\\d]{9,}$"
+        return passwordRegex.toRegex().matches(password)
+    }
```
philipplackner, line 42: For simple patterns like a valid password, I'm always a friend of doing this manually since it lets you quickly know what makes a valid password. Something like this: 
```
val hasLowerCase = password.any { it.isLowerCase() }
val hasUpperCase = password.any { it.isUpperCase() }
val hasDigit = password.any { it.hasDigit() }
val hasValidLength = password.length in min..max

return hasLowerCase && hasUpperCase && hasDigit && hasValidLength
```
That way, probably even my grandma could tell you what the requirements for a valid password are :D 
Regex makes sense if you can't easily do this manually, but has the cost of making code very unreadable.

philipplackner (approved):

PR #13, opened 2024-03-10
Started implementing the login functionality. I have the response com…

Mooky4591:
…ing back correctly but the wasLoginSuccessful boolean is always false sense it updates after it is already returned


app/src/main/java/com/example/taskyy/data/repositories/AuthRepositoryImpl.kt
```diff
@@ -1,43 +1,65 @@
...
+        }
+    }
+
+    @OptIn(DelicateCoroutinesApi::class)
+    override fun registerUser(user: User) {
+        val userDto = user.mapToUserDTO()
+        GlobalScope.launch(Dispatchers.IO) {
+            var response = retrofit.registerUser(userDto)
+        }
+    }
```
philipplackner, line 44: Be very careful with launching coroutine scopes outside of the ViewModel. Instead, expose functions as `suspend fun`. In this case, there is no way to track how long this `registerUser` call takes, since to the outside this function is not suspending as it launches a new coroutine which does the register call in parallel.
So instead, do this:
```
override suspend fun registerUser(user: User): Result<User> {
    val userDto = user.mapToUserDTO()
    val response = retrofit.registerUser(userDto)
    // Parse Response
    
    // If response successful
    return Result.success(user)
    
    // If response error
    return Result.error(throwable)
}
```
Then you can execute it in the ViewModel like this:
```
viewModelScope.launch {
    state = state.copy(isLoading = true)
    repository.registerUser(...)
        .onSuccess { user -> 
        
        }
        .onError { throwable -> 
        
        }
    state = state.copy(isLoading = false)
}
```

app/src/main/java/com/example/taskyy/data/repositories/AuthRepositoryImpl.kt
```diff
@@ -1,43 +1,65 @@
...
     override fun validatePassword(password: String): Boolean {
-        val passwordRegex = "^(?=.*[A-Za-z])(?=.*\\d)[A-Za-z\\d]{9,}$"
-        return passwordRegex.toRegex().matches(password)
+        val hasLowerCase = password.any { it.isLowerCase() }
+        val hasUpperCase = password.any { it.isUpperCase() }
+        val hasDigit = password.any { it.isDigit() }
+        val hasValidLength = password.length in 8..20
+
+        return hasLowerCase && hasUpperCase && hasDigit && hasValidLength
+    }
```
philipplackner, line 60: Validating the password is business logic and therefore responsibility of the domain layer. You can put this in a separate class for domain :)

app/src/main/java/com/example/taskyy/data/repositories/AuthRepositoryImpl.kt
```diff
@@ -1,43 +1,65 @@
...
+
+    override fun login(login: Login): Call<LoginUserResponse> {
```
philipplackner, line 62: Here, you're exposing a class from Retrofit (`Call`) to the domain layer which is not allowed. Domain shouldn't know which libraries you're using in the data layer, so it doesn't need to change when such a library is swapped out. 
Instead, I'd recommend to look into result wrappers, similar to what I shared in my comment above

app/src/main/java/com/example/taskyy/domain/usecases/LoginUseCase.kt
```diff
@@ -1,10 +1,42 @@
 package com.example.taskyy.domain.usecases
 
 import android.util.Patterns
+import com.example.taskyy.data.remote.LoginUserResponse
+import com.example.taskyy.domain.objects.Login
+import com.example.taskyy.domain.repository.AuthRepository
+import retrofit2.Call
+import retrofit2.Callback
+import retrofit2.Response
+import javax.inject.Inject
```
philipplackner, line 10: Here you can see how much your use case is coupled to other libraries, which is not allowed as described in my previous comment.
Instead, the repository should execute the Retrofit call and do the response handling. Then all the info the use case gets is whether logging in was successful or not (with potential data or errors)

philipplackner (approved):

PR #14, opened 2024-03-11
Attempting to change the login function so that the repository handle…

Mooky4591:
…s the login call and returns if it was successful or not. Having a hard time figuring out how to return the Result

### DIFF
--- a/app/src/main/java/com/example/taskyy/data/repositories/AuthRepositoryImpl.kt
+++ b/app/src/main/java/com/example/taskyy/data/repositories/AuthRepositoryImpl.kt
@@ -1,5 +1,6 @@
 package com.example.taskyy.data.repositories
 
+import android.util.Log
 import com.example.taskyy.data.local.data_access_objects.UserDao
 import com.example.taskyy.data.local.room_entity.UserEntity
 import com.example.taskyy.data.remote.LoginUserResponse
@@ -14,7 +15,12 @@ import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 import retrofit2.Call
+import retrofit2.Callback
+import retrofit2.Response
+import retrofit2.await
+import retrofit2.awaitResponse
 import javax.inject.Inject
+import kotlin.math.log
 
 class AuthRepositoryImpl @Inject constructor(
     private val userDao: UserDao,
@@ -22,30 +28,30 @@ class AuthRepositoryImpl @Inject constructor(
 ): AuthRepository {
 
     override fun addUserToDatabase(user: User) {
-        runBlocking { launch {
-            userDao.insertUser(user.mapToUserEntity())
-             }
-        }
-    }
-
-    override fun addTokenAndIdToDatabase(response: LoginUserResponse, email: String) {
-        runBlocking { launch {
-            userDao.update(response.token, response.userId, email)
+        runBlocking {
+            launch {
+                userDao.insertUser(user.mapToUserEntity())
             }
         }
     }
 
-    @OptIn(DelicateCoroutinesApi::class)
-    override fun registerUser(user: User) {
-        val userDto = user.mapToUserDTO()
-        GlobalScope.launch(Dispatchers.IO) {
-            var response = retrofit.registerUser(userDto)
+    override fun addTokenAndIdToDatabase(response: LoginUserResponse, email: String) {
+        runBlocking {
+            launch {
+                userDao.update(response.token, response.userId, email)
+            }
         }
+    }
+
+    override suspend fun registerUser(user: User) {
+        val userDto = user.mapToUserDTO()
+        retrofit.registerUser(userDto)
     }
 
     private fun User.mapToUserEntity(): UserEntity {
         return UserEntity(null, fullName, email, null)
     }
+
     private fun User.mapToUserDTO(): RegisterUserDTO {
         return RegisterUserDTO(fullName, email, password)
     }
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/example/taskyy/data/repositories/AuthRepositoryImpl.kt
```diff
@@ -14,38 +15,43 @@ import kotlinx.coroutines.GlobalScope
...
     override fun addUserToDatabase(user: User) {
-        runBlocking { launch {
-            userDao.insertUser(user.mapToUserEntity())
-             }
+        runBlocking {
+            launch {
+                userDao.insertUser(user.mapToUserEntity())
+            }
         }
     }
```
philipplackner: No runBlocking here as this can block your app's UI. If you have a suspend function like `userDao.insertUser` which you'd like to call, you have two options:
1. Call it in a launched coroutine (e.g. `viewModelScope`, `GlobalScope` or similar)
2. Call it in another suspend function

In the last PR I mentioned that I almost always recommend to expose suspend functions to your ViewModel and then execute them in `viewModelScope.launch` there. So you would need to change all your repo functions to `suspend fun`:
```
// Also add this suspend keyword to the AuthRepository interface
override suspend fun addUserToDatabase(user: User) {
    // Here you can now execute other suspend functions without runBlocking or a scope
}
```

app/src/main/java/com/example/taskyy/data/repositories/AuthRepositoryImpl.kt
```diff
@@ -59,7 +65,26 @@ class AuthRepositoryImpl @Inject constructor(
...
+                if(response.isSuccessful){
+                    returnStatement = Result.success()
+                } else {
+                    returnStatement = Result.failure(exception = Throwable())
+                }
+            }
+
+            override fun onFailure(call: Call<LoginUserResponse>, t: Throwable) {
+            }
+        })
```
philipplackner: You're using Retrofit's callback mechanism here, which is all fine but doesn't work well with coroutines. Instead, the easiest is to make all your Retrofit functions `suspend fun`:
```
@POST("/login)
suspend fun login(@Body body: LoginRequest): UserDto
```
Then you can call them in your repo just like this:
```
override suspend fun login(login: Login): Result<LoginUserResponse> {
    return try {
        val user = retrofit.login(...)
        Result.success(user.toUser())
    } catch(e: HttpException) {
        Result.failure(e)
    } catch(e: IOException) {
        Result.failure(e)
    }
```
I personally use a custom `Result` wrapper class which is a bit more flexible and optimized than the Kotlin one here, but at this point we should focus on you understanding the core mechanism behind this. Then we can later get into the little optimizations :)

app/src/main/java/com/example/taskyy/domain/repository/AuthRepository.kt
```diff
@@ -8,7 +8,7 @@ import retrofit2.Call
 interface AuthRepository {
     fun addUserToDatabase(user: User)
     fun addTokenAndIdToDatabase(response: LoginUserResponse, email: String)
-    fun registerUser(user: User)
+    suspend fun registerUser(user: User)
     fun validatePassword(password: String): Boolean
-    fun login(login: Login): Call<LoginUserResponse>
+    fun login(login: Login): Result<LoginUserResponse>
```
philipplackner: Becoming better 👏

app/src/main/java/com/example/taskyy/ui/viewmodels/LoginViewModel.kt
```diff
@@ -21,12 +24,20 @@ class LoginViewModel @Inject constructor(
...
             is LoginEvent.OnNameChanged -> state = state.copy(name = event.name)
         }
     }
+
+    fun login(event: Login){
+        state = state.copy(isLogginIn = true)
+        viewModelScope.launch {
+            state = state.copy(isLoginSuccessful = loginUseCase.loginUser(event))
+        }
+        state = state.copy(isLogginIn = false)
```
philipplackner: The first and last `state.copy` needs to go inside the `viewModelScope` block. The launched coroutine is running in parallel here independent of the "normal" execution flow. That means the code first updates `isLoggingIn` to true, then launches the coroutine (which happens instantly) and then updates `isLoggingIn` to false right after. But since the login call happens inside the coroutine, the function can't wait here until it finishes. That is only possible within the coroutine, so it would need to be like this:
```
viewModelScope.launch {
     state = state.copy(isLogginIn = true)
     state = state.copy(isLoginSuccessful = loginUseCase.loginUser(event))
     state = state.copy(isLogginIn = false) // Now this will be executed after loginUser
}
```

philipplackner (approved):

app/src/main/java/com/example/taskyy/data/repositories/AuthRepositoryImpl.kt
Mooky4591: This is confusing for me. Why would I return a UserDto (which currently doesn't exist) from the login request? Also, in the repo, why am I attempting to pass user.toUser() into the Result.success()? I don't understand what I'm doing here. I created the LoginResponse class which I assumed would be used as the return type from the login call

philipplackner: I think I commented on that earlier, but in the type of architecture you chose (presentation/domain/data layers), communication always points inwards at domain. That means both presentation and data layers are allowed to access classes from domain, but domain is not allowed to access any other layer. It also shouldn't know where data comes from because it would then indirectly know something about the data layer. 

That's why we have both data models (Such as `UserDto` from the API or `UserEntity` from the DB) which we map to a domain model in the repository. That way, domain only knows its own model and not any others from the data layer (such as your `LoginUserResponse` class).

What I called `UserDto` is the same as your `LoginUserResponse` here, we just used different naming. Either way, to adhere to the architectural guidelines, you would need to convert the `LoginUserResponse` to a model from domain.

PR #15, opened 2024-03-14
Created the Agenda Screen and got the date picker working. I also att…

Mooky4591:
…empted to implement the logout functionality from the agenda screen but I am getting a 401 error. Also I am not sure how to provide the current month to the Agenda screen to show the current month as the default month.

### DIFF
--- a/app/src/main/java/com/example/taskyy/data/repositories/AgendaRepositoryImpl.kt
+++ b/app/src/main/java/com/example/taskyy/data/repositories/AgendaRepositoryImpl.kt
@@ -1,0 +1,24 @@
+package com.example.taskyy.data.repositories
+
+import android.util.Log
+import com.example.taskyy.data.remote.TaskyyApi
+import com.example.taskyy.domain.repository.AgendaRepository
+import retrofit2.HttpException
+import javax.inject.Inject
+
+class AgendaRepositoryImpl @Inject constructor(
+    private val retrofit: TaskyyApi,
+    ): AgendaRepository {
+
+    override suspend fun logout(): Boolean {
+        var wasLogoutSuccessful = true
+        try {
+            retrofit.logoutUser()
+        } catch (e: HttpException) {
+            Log.e("Tag", e.code().toString())
+            wasLogoutSuccessful = false
+        }
+        return wasLogoutSuccessful
+    }
+
+}
--- a/app/src/main/java/com/example/taskyy/di/AuthModule.kt
+++ b/app/src/main/java/com/example/taskyy/di/AuthModule.kt
@@ -1,6 +1,7 @@
 package com.example.taskyy.di
 
 import android.content.Context
+import androidx.navigation.NavController
 import androidx.room.Room
 import com.example.taskyy.R
 import com.example.taskyy.data.local.data_access_objects.UserDao
@@ -23,7 +24,6 @@ import retrofit2.Retrofit
 import retrofit2.converter.moshi.MoshiConverterFactory
 import javax.inject.Singleton
 @Module
-
 @InstallIn(SingletonComponent::class)
 object AuthModule {
 
@@ -31,6 +31,12 @@ object AuthModule {
     @Singleton
     fun provideAuthRepository(userDao: UserDao, api: TaskyyApi): AuthRepository {
         return AuthRepositoryImpl(userDao, api)
+    }
+
+    @Provides
+    @Singleton
+    fun provideNavController(@ApplicationContext context: Context): NavController {
+        return NavController(context = context)
     }
 
     @Provides
--- a/app/src/main/java/com/example/taskyy/ui/AgendaScreen.kt
+++ b/app/src/main/java/com/example/taskyy/ui/AgendaScreen.kt
@@ -1,12 +1,203 @@
 package com.example.taskyy.ui
 
+import androidx.compose.foundation.background
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.fillMaxHeight
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.shape.AbsoluteRoundedCornerShape
+import androidx.compose.foundation.shape.CircleShape
+import androidx.compose.material3.DatePicker
+import androidx.compose.material3.DatePickerDefaults
+import androidx.compose.material3.DisplayMode
+import androidx.compose.material3.DropdownMenu
+import androidx.compose.material3.DropdownMenuItem
+import androidx.compose.material3.ExperimentalMaterial3Api
+import androidx.compose.material3.Icon
+import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
+import androidx.compose.material3.rememberDatePickerState
 import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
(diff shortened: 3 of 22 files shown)


app/src/main/java/com/example/taskyy/data/repositories/AgendaRepositoryImpl.kt
```diff
@@ -0,0 +1,24 @@
...
+
+    override suspend fun logout(): Boolean {
+        var wasLogoutSuccessful = true
+        try {
+            retrofit.logoutUser()
+        } catch (e: HttpException) {
+            Log.e("Tag", e.code().toString())
+            wasLogoutSuccessful = false
+        }
+        return wasLogoutSuccessful
```
philipplackner, line 21: You can also use try/catch as an expression and return it :)
```
return try {
    retrofit.logoutUser()
    true
} catch(e: HttpException) {
    false
}
```
Also, with Retrofit it's important to catch `IOException` in addition to `HttpException`. They're thrown when there's no internet connection for example or when the JSON parsing fails

app/src/main/java/com/example/taskyy/di/AuthModule.kt
```diff
@@ -33,6 +33,12 @@ object AuthModule {
         return AuthRepositoryImpl(userDao, api)
     }
 
+    @Provides
+    @Singleton
+    fun provideNavController(@ApplicationContext context: Context): NavController {
+        return NavController(context = context)
+    }
```
philipplackner, line 40: I don't think this is the proper way to create a NavController for Compose. I would always prefer `rememberNavController` since that's aware of the current composable and will be properly destroyed when not needed anymore

app/src/main/java/com/example/taskyy/ui/AgendaScreen.kt
```diff
@@ -1,12 +1,203 @@
...
+import com.example.taskyy.ui.events.AgendaEvent
+import com.example.taskyy.ui.viewmodels.AgendaState
+import java.time.LocalDateTime
+import java.time.ZoneId
 
 @Composable
-fun AgendaScreen(){
-    Column {
-        Text(text = "Agenda Screen")
+fun AgendaScreen(state: AgendaState, navController: NavController, onEvent: (AgendaEvent) -> Unit){
```
philipplackner, line 46: A NavController reference can also quickly break your Compose preview which is why we usually keep that outside the screen composables. Since you already provide an `onEvent` lambda here (which is good!), you can listen to these events in the `composable` block of your NavHost and navigate there:
```
composable("agenda) {
    AgendaScreen(
        state = viewModel.state,
        onEvent = { event ->
            when(event) {
                is AgendaEvent.Logout -> // navigate to login
                is AgendaEvent.OnAgendaItemClick -> // navigate to agenda details
                ...
            }
            viewModel.onEvent(event) // Still send all events to the VM so nothing is lost
        }
    )
}
```

app/src/main/java/com/example/taskyy/ui/AgendaScreen.kt
```diff
@@ -1,12 +1,203 @@
...
+            CircleWithInitials(
+                isUserDropDownExpanded = state.isUserDropDownExpanded,
+                onUserInitialsClicked = { isUserDropDownExpanded ->
+                    onEvent(AgendaEvent.OnUserInitialsClicked(isUserDropDownExpanded))
+                },
+                onItemSelected = {
+                    onEvent(AgendaEvent.OnLogOutCLicked)
+                },
+            )
+        }
```
philipplackner, line 86: Your screen is quite big here which is why I'd recommend to split it up into more smaller composables. This piece here could be the `AgendaToolbar` for example. Separating a screen by such sections makes it much easier to scan its content

app/src/main/java/com/example/taskyy/ui/AgendaScreen.kt
```diff
@@ -1,12 +1,203 @@
...
+                    .fillMaxHeight(.75f)
+                    .padding(30.dp)
+            ) {
+                Spacer(modifier = Modifier.height(10.dp))
+            }
+        }
     }
+    if(state.wasLogoutSuccessful){
+        navController.navigate(Screen.Login.route)
+    }
```
philipplackner, line 106: Something like this is dangerous - we call that a side-effect. Since you have no control over when and how often a composable function is called, putting non-Compose functions (like `navController.navigate`) inside a composable function can lead to very weird behavior. In this case it's unlikely going to lead to an issue because `state.wasLogoutSuccessful` is false for most of the time, but in general, you'd want to handle these with Compose's effect handlers. In this case like this:
```
// This block will trigger every time the passed state changes
LaunchedEffect(state.wasLogoutSuccessful) {
    if(state.wasLogoutSuccessful) {
        navController.navigate(Screen.Login.route) {
            popUpTo(Screen.Agenda.route) { // This is important, since otherwise the user can get back to the agenda screen after logging out
                inclusive = true 
            }
        }
    }
}
```

app/src/main/java/com/example/taskyy/ui/LoginScreen.kt
```diff
@@ -80,6 +80,9 @@ fun LoginScreen(state: LoginState, onEvent: (LoginEvent) -> Unit, navController:
             }
         }
     }
+    if(state.isLoginSuccessful){
+        navController.navigate(Screen.Agenda.route)
+    }
```
philipplackner, line 85: Same here:
1. Put this in `LaunchedEffect`
2. Pop the login screen from the backstack, so you can't get back to it after logging in

app/src/main/java/com/example/taskyy/ui/viewmodels/AgendaViewModel.kt
```diff
@@ -0,0 +1,53 @@
...
+            is AgendaEvent.OnMonthExpanded -> state = state.copy(isMonthExpanded = event.isMonthExpanded)
+            is AgendaEvent.OnDateSelected -> formatSelectedDate(event.date)
+            is AgendaEvent.OnUserInitialsClicked -> state = state.copy(isUserDropDownExpanded = event.isUserDropDownExpanded)
+            is AgendaEvent.OnLogOutCLicked -> logout()
+        }
+    }
+
+    private fun formatSelectedDate(date: Long) {
+        val monthDate = SimpleDateFormat("MMMM", Locale.US)
+        val monthName = monthDate.format(date)
```
philipplackner, line 34: Generally, I'd recommend to use `DateTimeFormatter` for date formatting. `SimpleDateFormat` is an older API which isn't thread-safe. In your example this won't lead to any issues, since your instance isn't shared across threads, but to build a habit, I'd start using the more modern API:
```
val now = LocalDateTime.now()
val formattedTime = DateTimeFormatter
    .ofPattern("dd MM yyyy")
    .format(now)
```
Generally, the `LocalDateTime` API is a lot more flexible and allows easy date time calculations

philipplackner (approved):

app/src/main/java/com/example/taskyy/ui/AgendaScreen.kt
Mooky4591: If I do not pass the navController into the agenda screen and I am handling navigation from the nav host, would I put this in the navhost since the agenda screen composable wouldn't have a navController reference?

philipplackner: Yes, or in the NavHost you could simply do what I described in another comment:
```
composable("agenda) {
    AgendaScreen(
        state = viewModel.state,
        onEvent = { event ->
            when(event) {
                is AgendaEvent.Logout -> // navigate to login
                is AgendaEvent.OnAgendaItemClick -> // navigate to agenda details
                ...
            }
            viewModel.onEvent(event) // Still send all events to the VM so nothing is lost
        }
    )
}
```

Mooky4591: When I implement it this way, it no longer validates the log in. It just logs the user in. I added

`is AgendaEvent.Login -> if(state.isLoginSuccessful) {
          navController.navigate(Screen.Agenda.route)
          }`

This makes me have to click login twice since the state isn't updated immediately on validation.

philipplackner: In how far does the agenda feature have something to do with logging in? 🤔

PR #16, opened 2024-03-15
Refactored the Login, Register, and Agenda screens so that the naviga…

Mooky4591:
…tion is handled in the navhost. I also added the add agenda item button to the bottom of the screen. I know I am doing that incorrectly but I'm not sure how to get it down to the corner. I also am still getting a 401 error when attempting to logout and I'm not entirely sure why that is

### DIFF
--- a/app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
+++ b/app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
@@ -9,6 +9,9 @@ import androidx.navigation.navigation
 import com.example.taskyy.ui.AgendaScreen
 import com.example.taskyy.ui.LoginScreen
 import com.example.taskyy.ui.RegisterScreen
+import com.example.taskyy.ui.events.AgendaEvent
+import com.example.taskyy.ui.events.LoginEvent
+import com.example.taskyy.ui.events.RegisterEvent
 import com.example.taskyy.ui.viewmodels.AgendaViewModel
 import com.example.taskyy.ui.viewmodels.LoginViewModel
 import com.example.taskyy.ui.viewmodels.RegisterViewModel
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
```diff
@@ -26,24 +29,53 @@ fun Nav() {
...
                 val loginViewModel = hiltViewModel<LoginViewModel>()
                 val state = loginViewModel.state
-                LoginScreen(state, loginViewModel::onEvent, navController)
+                LoginScreen(
+                    state = state,
+                    onEvent = { event ->
+                        when (event) {
+                            is LoginEvent.OnLoginClick -> if (state.isLoginSuccessful) {
+                                (navController.navigate(Screen.Agenda.route))
+                            }
```
philipplackner, line 38: For logging in it's a bit different. Here you only want to navigate when logging was successful and not generally after clicking the button. When you click the button for the first time here, the login wasn't successful yet, so nothing will happen.
Instead, you need to work with events the ViewModel sends to the UI. You can do that with a Channel:
```
class LoginViewModel: ViewModel() {

    private val eventChannel = Channel<LoginEvent>() // Different set of events
    val events = eventChannel.receiveAsFlow()
    
    fun login() {
        // Do login...
        
        if(loginSuccess) {
            eventChannel.send(LoginEvent.LoginSuccess)
        }
    }
}
```
Then you can use this helper composable in your composable block in the nav host to listen to these events and do the navigation there:
```
@Composable
private fun <T> ObserveAsEvents(flow: Flow<T>, onEvent: (T) -> Unit) {
    val lifecycleOwner = LocalLifecycleOwner.current
    LaunchedEffect(flow, lifecycleOwner.lifecycle) {
        lifecycleOwner.repeatOnLifecycle(Lifecycle.State.STARTED) {
            withContext(Dispatchers.Main.immediate) {
                flow.collect(onEvent)
            }
        }
    }
}
```
In your Navhost:
```
composable("login") {
    val viewModel = hiltViewModel<LoginViewModel>()
    ObserveAsEvents(viewModel.events) { event ->
        when(event) {
            is LoginEvent.Login -> // Navigate here
        }
    }
}
```
So to summarize:
Instant navigation after button click? -> You can navigate in the UI directly
Navigation only after an asynchronous call in the VM? -> Navigate with via Channel as above

app/src/main/java/com/example/taskyy/ui/AgendaScreen.kt
```diff
@@ -86,23 +79,60 @@ fun AgendaScreen(state: AgendaState, navController: NavController, onEvent: (Age
...
+                //I know this is wrong but I don't know how to push this button to the bottom right of the screen
+                Spacer(modifier = Modifier.height(645.dp))
+                Row(
+                    horizontalArrangement = Arrangement.End,
+                    modifier = Modifier
+                        .fillMaxSize()
+                        .padding(horizontal = 10.dp)
+                ) {
+                    AddAgendaItem(onEvent = onEvent)
+                }
```
philipplackner, line 98: Such a button like the add button is called a floating action button. Best for that is to use a Scaffold:
```
Scaffold(
    floatingActionButton = {
        AddAgendaItem(...)
    }
) { padding -> 
    // Screen content
}
```

philipplackner (approved):

PR #17, opened 2024-03-15
Refactored the agenda screen to use a scaffold. The white shape no lo…

Mooky4591:
…nger goes to the bottom of the screen. I'm not entirely sure why. I also implemented the eventChannel for login and register. It now takes a moment to log in on the click so I will likely need to implement a spinner. I am still getting the error on attempting to logout though. I also implemented the logic for displaying the dates across the screen when selecting a day from the date picker.

questions:
1. I am still getting a 400 error when attempting to log out. I am not sure if I did something wrong with the implementation of the interceptor.
2. I am using a surface for the white design on the screen and after implementing the scaffold on the agenda screen, it no longer reaches the bottom of the screen and I'm not entirely sure why.
3. When it displays the calendar after selecting the month in the top corner, the calendar is pushed off the edge of the screen. I'm not sure if there is a way to correct that

### DIFF
--- a/app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
+++ b/app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
@@ -1,7 +1,13 @@
 package com.example.taskyy.domain.navigation
 
+import android.os.Build
+import androidx.annotation.RequiresApi
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.ui.platform.LocalLifecycleOwner
 import androidx.hilt.navigation.compose.hiltViewModel
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.repeatOnLifecycle
 import androidx.navigation.compose.NavHost
 import androidx.navigation.compose.composable
 import androidx.navigation.compose.rememberNavController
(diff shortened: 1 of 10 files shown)


app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
```diff
@@ -15,28 +21,48 @@ import com.example.taskyy.ui.events.RegisterEvent
...
+    fun <T> ObserveAsEvents(flow: Flow<T>, onEvent: (T) -> Unit) {
+        val lifecycleOwner = LocalLifecycleOwner.current
+        LaunchedEffect(flow, lifecycleOwner.lifecycle) {
+            lifecycleOwner.repeatOnLifecycle(Lifecycle.State.STARTED) {
+                withContext(Dispatchers.Main.immediate) {
+                    flow.collect(onEvent)
+                }
+            }
+        }
+    }
```
philipplackner, line 45: To keep it reusable it makes sense to define this function in a normal separate file

app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
```diff
@@ -46,14 +72,19 @@ fun Nav() {
...
+                        else -> {}
+                    }
+                }
                 val state = registerViewModel.state
                 RegisterScreen(
                     state = state,
                     onEvent = { event ->
                         when (event) {
-                            is RegisterEvent.OnGetStartedClick -> if (state.isRegistrationSuccessful) {
+                            is RegisterEvent.OnGetStartedClick -> if (state.isRegistrationSuccessful)
```
philipplackner, line 86: This is not needed anymore since you handle the navigation in `ObserveAsEvents`

app/src/main/java/com/example/taskyy/ui/AgendaScreen.kt
```diff
@@ -42,77 +45,141 @@ import com.example.taskyy.ui.viewmodels.AgendaState
...
-            )
-        }
-        Surface(
-            modifier = Modifier
-                .fillMaxHeight(.90F)
-                .fillMaxWidth(),
-            shape = AbsoluteRoundedCornerShape(30.dp, 30.dp)
-        ) {
+        },
+        content = { PaddingValues ->
```
philipplackner, line 88: This padding should normally be applied to the root composable of the Scaffold, so in this case your Column. That will apply the necessary padding from the Scaffold, depending on which components you added to it. You added a Toolbar here, so the padding will push the real content to the bottom by the toolbar's height, so nothing overlaps

app/src/main/java/com/example/taskyy/ui/AgendaScreen.kt
```diff
@@ -42,77 +45,141 @@ import com.example.taskyy.ui.viewmodels.AgendaState
...
+                                        horizontalAlignment = Alignment.CenterHorizontally
+                                    ) {
+                                        SelectedDaysOfTheWeek(dayOfTheWeek = day.dayOfTheWeek)
+                                        Spacer(modifier = Modifier.height(10.dp))
+                                        SelectedDaysOfTheMonth(dayOfTheMonth = day.dayOfTheMonth)
+                                    }
+                                }
+                            }
+                        }
+                    }
```
philipplackner, line 124: I'd really recommend to work more with modular components for your screen, since otherwise your screen quickly bloats up and it becomes a struggle to know what to change where. This could simply be extracted into a separate composable that describes what it contains

app/src/main/java/com/example/taskyy/ui/AgendaScreen.kt
```diff
@@ -42,77 +45,141 @@ import com.example.taskyy.ui.viewmodels.AgendaState
...
+        },
+    )
+}
+
+@Composable
+fun SelectedDaysOfTheMonth(dayOfTheMonth: String) {
+    Text(text = dayOfTheMonth, fontWeight = FontWeight.Bold, fontSize = 15.sp)
+}
+
+@RequiresApi(Build.VERSION_CODES.Q)
```
philipplackner, line 139: Why does this require Android Q?

app/src/main/java/com/example/taskyy/ui/viewmodels/AgendaViewModel.kt
```diff
@@ -25,21 +26,62 @@ class AgendaViewModel @Inject constructor(
...
+            )
+            dayList.add(day)
+        }
+
+
+        state = state.copy(
+            selectedMonth = monthFormatter.format(localDateTime).uppercase(),
+            selectedDay = dayList,
+        )
+    }
```
philipplackner, line 75: I'm very confused about what this function does 😅  Why do you need 3 ArrayLists? 
Apparently the only goal is to format the month and days list. You first construct a temp date list and a day list, just to then construct a new temporary dateList which gets then assigned to the dayList. Why not populate the dayList right away?
To construct the complete day list, all you need is this:
```
val days = (0..6).map {
    val date = localDateTime.plusDays(i.toLong()) 
    Day(
        dayOfTheWeek = dayOfTheWeekFormatter.format(date),
        dayOfTheMonth = dayOfTheMonthFormatter.format(date)
    )
}
```

philipplackner (approved):

PR #18, opened 2024-03-18
Had to fix a log in issue where a user couldn't log in a second time.…

Mooky4591:
… Also refactored the AgendScreen to try and make it more modular. Moved the screens into a screen package. I got the highlight working when user selects a dat from the date array on the agenda screen.

### DIFF
--- a/app/src/main/java/com/example/taskyy/data/repositories/AuthRepositoryImpl.kt
+++ b/app/src/main/java/com/example/taskyy/data/repositories/AuthRepositoryImpl.kt
@@ -2,9 +2,9 @@ package com.example.taskyy.data.repositories
 
 import com.example.taskyy.data.local.data_access_objects.UserDao
 import com.example.taskyy.data.local.room_entity.UserEntity
-import com.example.taskyy.data.remote.response_objects.LoginUserResponse
 import com.example.taskyy.data.remote.TaskyyApi
 import com.example.taskyy.data.remote.data_transfer_objects.RegisterUserDTO
+import com.example.taskyy.data.remote.response_objects.LoginUserResponse
 import com.example.taskyy.domain.objects.Login
 import com.example.taskyy.domain.objects.User
 import com.example.taskyy.domain.repository.AuthRepository
@@ -17,31 +17,20 @@ class AuthRepositoryImpl @Inject constructor(
     private val retrofit: TaskyyApi,
 ): AuthRepository {
 
-    override suspend fun addUserToDatabase(user: User) {
-        userDao.insertUser(user.mapToUserEntity())
-    }
-
     override suspend fun addTokenAndIdToDatabase(token: String, userId: String, email: String) {
         userDao.update(token, userId, email)
     }
 
     override suspend fun registerUser(user: User): Boolean {
-        val userDto = user.mapToUserDTO()
-        var wasRegistrationSuccessful = true
-        try {
+        val userDto = user.toUserDTO()
+        return try {
             retrofit.registerUser(userDto)
+            true
         } catch (e: HttpException) {
-            wasRegistrationSuccessful = false
+            false
+        } catch (e: IOException) {
+            false
         }
-        return wasRegistrationSuccessful
-    }
-
-    private fun User.mapToUserEntity(): UserEntity {
-        return UserEntity(null, fullName, email, null)
-    }
-
-    private fun User.mapToUserDTO(): RegisterUserDTO {
-        return RegisterUserDTO(fullName, email, password)
     }
 
     override fun validatePassword(password: String): Boolean {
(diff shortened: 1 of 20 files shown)


app/src/main/java/com/example/taskyy/data/repositories/AuthRepositoryImpl.kt
```diff
@@ -56,17 +45,31 @@ class AuthRepositoryImpl @Inject constructor(
     override suspend fun login(login: Login): Result<User> {
         return try {
             val user = retrofit.loginUser(login)
+
+            if (!userDao.doesUserExist(user.userId)) {
+                val userEntity = user.toUserEntity(login.email, user.userId)
+                userDao.insertUser(userEntity)
+            }
```
philipplackner, line 52: To avoid needing to check if the user exists, you can also annotate your insert function with `@Upsert`. That will insert a new entry or update the existing one if there is already one with the ID you're trying to insert

app/src/main/java/com/example/taskyy/di/AgendaModule.kt
```diff
@@ -26,7 +27,13 @@ object AgendaModule {
...
     @Singleton
     fun providesAgendaRepo(@ApplicationContext context: Context): AgendaRepository {
-        return AgendaRepositoryImpl(provideRetrofitInstance(context))
+        return AgendaRepositoryImpl(provideRetrofitInstance(context), provideUserDao(context))
+    }
+
+    @Provides
+    @Singleton
+    fun provideUserDao(@ApplicationContext context: Context): UserDao {
+        return AuthModule.provideRoomDatabase(context).userDao()
```
philipplackner, line 36: Don't call these provide functions yourself, that completely breaks their lifecycle. Every time you call `provideRoomDatabase`, you'll create a new instance of it. Let Dagger handle this instead and don't call these yourself. To do that, just pass the dependencies as parameters here and Dagger will insert them:
```
fun provideUserDao(@ApplicationContext context: Context, db: TaskyyDatabase): UserDao {
    return db.userDao()
}
```

app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
```diff
@@ -72,40 +55,38 @@ fun Nav() {
...
+                arguments = listOf(
+                    navArgument("email") { type = NavType.StringType }
+                )
+            ) { backStackEntry ->
+                val email: String = backStackEntry.arguments?.getString("email") ?: ""
                 val agendaViewModel = hiltViewModel<AgendaViewModel>()
                 val state = agendaViewModel.state
+                if (email != null) {
+                    agendaViewModel.setUserInitials(email)
+                }
```
philipplackner, line 80: This is a side-effect in Compose, since you execute non-Compose code (`agendaViewModel.setUserInitials()`) in a Composable context. That's not ideal, since you can't control when and how often a Compose function is called (that's handled by the Compose runtime). Having such non-Compose functions in there could execute them more often than you want to.
In this specific case it's unlikely going to be an issue, but imagine this call to the VM would fire a network request. Then every time, there's a recomposition (potentially many times a second if you're running an animation), you'd fire another network request.

Instead, you can get the email argument directly in your ViewModel with SavedStateHandle:
```
@HiltViewModel
class AgendaViewModel @Inject constructor(
    // ...
    savedStateHandle: SavedStateHandle // Will be injected automatically by Hilt
): ViewModel() {

    private val email = savedStateHandle.get<String>("email")
}
```

app/src/main/java/com/example/taskyy/ui/viewmodels/AgendaViewModel.kt
```diff
@@ -53,53 +66,82 @@ class AgendaViewModel @Inject constructor(
...
+    }
+
+    fun setUserInitials(email: String) {
+        viewModelScope.launch {
+            val name = agendaRepository.getUserName(email)
+            state = state.copy(name = name)
+            state = state.copy(initials = name
+                .split(' ')
+                .mapNotNull { it.firstOrNull()?.toString() }
+                .reduce { acc, s -> acc + s })
```
philipplackner, line 108: If there's a middle name, you'll get 3 initials here. In that case, the expected behavior would probably taking the initials of the first and last name ignoring the middle name

philipplackner (approved):

app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
Mooky4591: I am getting a little confused with all the different state options. For some variables I can set using state = state.copy.... others need state flow and now I have another variable that should be set using savedStateHandle. Is there a cheat sheet or some way to remember what should be set how? Or maybe there is  a standard way of doing this that works for all variables? I worry I won't be able to keep it straight

philipplackner: Compose state and StateFlow both have the same purpose: To hold state and update the UI when it changes. Using StateFlow in the VM requires converting it into a format Compose understands by using `collectAsState()` in your UI. If you use Compose state in the VM directly, that's not needed. You can therefore completely stick to Compose state in your project and not worry about StateFlow at all. 

SavedStateHandle is just a bundle that is passed to each screen. You can use it to retrieve navigation arguments in the ViewModel, like your email here. But since these arguments are static and don't change, they're not really considered state (state = a value that can change over time)

Mooky4591: Do I need to put these navigation arguments somewhere or does saveStateHandle take care of all of that? Also, I used that to get the email but I can't call my viewmodel method from the viewmodel unless I put It in the onEvent function. I can't call it from a composable so how do I launch the method to create the initials?

philipplackner: You don't need to put it there, that's all handled by the navigation framework already

PR #19, opened 2024-03-20
03 18 24 error handling

Mooky4591:


app/src/main/java/com/example/taskyy/ui/viewmodels/LoginViewModel.kt
```diff
@@ -40,23 +40,31 @@ class LoginViewModel @Inject constructor(
...
+                        eventChannel.send(LoginEvent.LoginSuccess(state.email))
+                    }
 
-                if (state.isLoginSuccessful) {
-                    eventChannel.send(LoginEvent.LoginSuccess(state.email))
+                    is Result.Error -> {
+                        //I can't parse this here because the asString function requires a context. Where should this go?
+                        //state = state.copy(loginErrorMessage = login.error.asUiText().asString())
+                        state = state.copy(isLogginIn = false)
+                        eventChannel.send(LoginEvent.LoginFailed)
```
philipplackner, line 61: Your `loginErrorMessage` state should be of type `UiText`, so you can assign it here and unwrap it in the UI where you have the context

philipplackner (approved):

PR #20, opened 2024-03-21
Attempting to make changes requested. I don't have savedStateHandle w…

Mooky4591:
…orking correctly. It is showing as null. I tried manually adding the value in the login screen but it still shows as null


app/src/main/java/com/example/taskyy/ui/screens/AgendaScreen.kt
```diff
@@ -111,6 +118,10 @@ fun AgendaScreen(state: AgendaState, onEvent: (AgendaEvent) -> Unit) {
     )
 }
 
+fun setUserInitials(onEvent: () -> Unit) {
+    onEvent()
+}
```
philipplackner, line 123: This function seems pretty pointless to me :D

philipplackner (approved):

PR #21, opened 2024-03-23
Built out the reminder screen. I haven't added any real functionality…

Mooky4591:
… yet. I am worried that the way I am designing the screens isn't going to look good on larger devices. Any feed back would be appreciated

### DIFF
--- a/app/src/main/java/com/example/taskyy/ui/screens/ReminderScreen.kt
+++ b/app/src/main/java/com/example/taskyy/ui/screens/ReminderScreen.kt
@@ -1,0 +1,332 @@
+package com.example.taskyy.ui.screens
+
+import androidx.compose.foundation.Image
+import androidx.compose.foundation.background
+import androidx.compose.foundation.border
+import androidx.compose.foundation.clickable
+import androidx.compose.foundation.layout.Arrangement
+import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
+import androidx.compose.foundation.layout.Row
+import androidx.compose.foundation.layout.fillMaxHeight
+import androidx.compose.foundation.layout.fillMaxSize
+import androidx.compose.foundation.layout.fillMaxWidth
+import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.padding
+import androidx.compose.foundation.layout.size
+import androidx.compose.foundation.shape.AbsoluteRoundedCornerShape
+import androidx.compose.foundation.shape.CircleShape
+import androidx.compose.material3.HorizontalDivider
+import androidx.compose.material3.Scaffold
+import androidx.compose.material3.Surface
+import androidx.compose.material3.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.ColorFilter
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.unit.LayoutDirection
+import androidx.compose.ui.unit.dp
+import androidx.compose.ui.unit.sp
+import com.example.taskyy.R
+import com.example.taskyy.ui.events.ReminderEvent
+import com.example.taskyy.ui.viewmodels.ReminderState
+
+@Composable
+fun ReminderScreen(state: ReminderState, onEvent: (ReminderEvent) -> Unit) {
+    Scaffold(
+        content = {
+            Column(
+                verticalArrangement = Arrangement.SpaceEvenly,
+                modifier = Modifier
+                    .fillMaxSize()
+                    .background(Color.Black)
+                    .padding(it.calculateLeftPadding(layoutDirection = LayoutDirection.Ltr))
+            ) {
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/example/taskyy/ui/screens/ReminderScreen.kt
```diff
@@ -0,0 +1,332 @@
...
+                ) {
+                    Body(itemBoxTitle = "", reminderText = "", dateString = state.dateString)
+                }
+            }
+        }
+    )
+}
+
+@Composable
+fun Body(itemBoxTitle: String, reminderText: String, dateString: String) {
```
philipplackner, line 64: I think this name could be more descriptive, such as `ReminderScreenContent` or so. Also, you'll only need it in this file, so I'd make this function private

app/src/main/java/com/example/taskyy/ui/screens/ReminderScreen.kt
```diff
@@ -0,0 +1,332 @@
...
+        ItemBoxTitle("Reminder")
+    }
+    Column {
+        NewItemRow()
+        RowDivider(30)
+        ItemDescription()
+        RowDivider(25)
+        TimeAndDateRow(dateString = dateString)
+        RowDivider(topPadding = 25)
+        Alarm()
```
philipplackner, line 80: `Alarm` could also mean anything. I think you mean something like `ReminderTypeSection`, do you? 😄

app/src/main/java/com/example/taskyy/ui/screens/ReminderScreen.kt
```diff
@@ -0,0 +1,332 @@
...
+}
+
+@Composable
+fun Alarm() {
+    Row(
+        modifier =
+        Modifier
+            .padding(start = 20.dp, top = 20.dp)
+            .clickable {
+                //navigate to alarm selection screen
```
philipplackner, line 91: The selection should just be a drop down

app/src/main/java/com/example/taskyy/ui/screens/ReminderScreen.kt
```diff
@@ -0,0 +1,332 @@
...
+                .height(1.dp) // Adjust the height of the divider as needed
+                .padding(
+                    bottom = 4.dp,
+                    start = 20.dp,
+                    end = 20.dp
+                ) // Adjust the padding to control the distance between the divider and the content
+        ) {
+            HorizontalDivider(color = Color.LightGray) // Divider with black color
+        }
+    }
```
philipplackner, line 188: I think this is a bit overcomplicated for a simple divider. Something like this should do the same:
```
HorizontalDivider(
    color = Color.LightGray,
    modifier = Modifier
        .padding(
            horizontal = 20.dp,
            vertical = 4.dp
        )
)
```

app/src/main/java/com/example/taskyy/ui/screens/ReminderScreen.kt
```diff
@@ -0,0 +1,332 @@
...
+        modifier = Modifier
+            .padding(top = topPadding.dp)
+            .fillMaxWidth()
+    ) {
+        Image(
+            painter = painterResource(id = R.drawable.chevron_forward),
+            contentDescription = "proceed forward arrow",
+            modifier = Modifier.padding(end = 15.dp)
+        )
+    }
```
philipplackner, line 226: Also here, I'd expect that `ChevronForward` is a standalone icon and not a full row which spans the whole width and the icon is at the end. That keeps it less flexible since you force how this icon is used

app/src/main/java/com/example/taskyy/ui/screens/ReminderScreen.kt
```diff
@@ -0,0 +1,332 @@
...
+}
+
+@Composable
+fun AlarmIconWithBackground() {
+    Box(
+        modifier = Modifier
+            .size(32.dp) // Adjust the size of the vector asset as needed
```
philipplackner, line 233: I think this size should be set for the `Image` composable and not the container. If the size of the vector asset is smaller than 32dp it will also be displayed as a smaller icon, since you just force the size of its container

app/src/main/java/com/example/taskyy/ui/screens/ReminderScreen.kt
```diff
@@ -0,0 +1,332 @@
...
+        contentAlignment = Alignment.Center
+
+    ) {
+        Image(
+            painter = painterResource(
+                id = R.drawable.alarm_icon
+            ),
+            contentDescription = "Alarm icon",
+            colorFilter = ColorFilter.tint(color = Color(android.graphics.Color.parseColor("#aeb4bf")))
+        )
```
philipplackner, line 247: Also for icons, always use `Icon` instead of `Image`. That way, you can easily choose a tint (without any color filter magic):
```
Icon(
    imageVector = ImageVector.vectorResource(R.drawable.alarm_icon),
    tint = Color(0xFFAEB4BF)
)
```

philipplackner (approved):
It seems to me that you're hardcoding a lot of size values which can be what you want but sometimes that makes the UI less flexible when it should work for multiple screen sizes. 
It's hard to tell just from a PR how the UI would look on different screens, which is why it's best you try it out on a tablet emulator for example. It's also a big help for me if you could attach screenshots of the UI you've built to such PRs so I know how everything looks like visually

PR #22, opened 2024-03-24
Attempting to get savedStateHandle to work but I can't seem to get da…

Mooky4591:
…ta out of it. I am wanting to use it to pass data from screen to screen and use sharedPreferences for data that needs to persist across the app. Does this approach seem correct? Also, I don't know why I can't get data out of the savedStateHandle

### DIFF
--- a/app/src/main/java/com/example/taskyy/ui/events/ReminderEvent.kt
+++ b/app/src/main/java/com/example/taskyy/ui/events/ReminderEvent.kt
@@ -3,4 +3,14 @@ package com.example.taskyy.ui.events
 interface ReminderEvent {
 
     data object SetDateString : ReminderEvent
+    data class AlarmTypeDropDownSelected(var isAlarmSelectionExpanded: Boolean) : ReminderEvent
+    data class TimePickerSelected(var timePickerSelected: Boolean) : ReminderEvent
+    data class AlarmTimeTextSelected(var alarmTimeText: String) : ReminderEvent
+    data class TimeSelected(var selectedTime: String) : ReminderEvent
+    data class ReminderDescriptionUpdated(var reminderDescription: String) : ReminderEvent
+    data class ReminderTitleTextUpdated(var reminderTitleText: String) : ReminderEvent
+    data class UpdateDateSelection(var selectedDate: Long) : ReminderEvent
+    data class DatePickerSelcted(var datePickerExpanded: Boolean) : ReminderEvent
+    data object SaveSelected : ReminderEvent
+    data object Close : ReminderEvent
 }
--- a/app/src/main/java/com/example/taskyy/ui/screens/ReminderScreen.kt
+++ b/app/src/main/java/com/example/taskyy/ui/screens/ReminderScreen.kt
@@ -16,7 +16,10 @@ import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.shape.AbsoluteRoundedCornerShape
 import androidx.compose.foundation.shape.CircleShape
+import androidx.compose.material3.DropdownMenu
+import androidx.compose.material3.DropdownMenuItem
 import androidx.compose.material3.HorizontalDivider
+import androidx.compose.material3.Icon
 import androidx.compose.material3.Scaffold
 import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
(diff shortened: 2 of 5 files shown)


app/src/main/java/com/example/taskyy/ui/events/ReminderEvent.kt
```diff
@@ -3,4 +3,14 @@ package com.example.taskyy.ui.events
 interface ReminderEvent {
 
     data object SetDateString : ReminderEvent
+    data class AlarmTypeDropDownSelected(var isAlarmSelectionExpanded: Boolean) : ReminderEvent
+    data class TimePickerSelected(var timePickerSelected: Boolean) : ReminderEvent
+    data class AlarmTimeTextSelected(var alarmTimeText: String) : ReminderEvent
```
philipplackner, line 8: Could make sense to make use of an enum class here to have a set of distinct options for these reminders:
```
enum class ReminderType {
    10_MINUTES_BEFORE,
    30_MINUTES_BEFORE,
    ...
}
```
That also makes it easier to check what was passed in the VM, since a when-expression forces you to go through all the different options

app/src/main/java/com/example/taskyy/ui/screens/ReminderScreen.kt
```diff
@@ -71,60 +83,113 @@ fun Body(itemBoxTitle: String, reminderText: String, dateString: String) {
...
+            })
     }
 }
 
 @Composable
-fun Alarm() {
+fun SetAlarmTimeRow(
+    isAlarmSelectionExpanded: Boolean,
+    alarmTimeSelectionText: String,
+    onEvent: (Boolean) -> Unit
```
philipplackner, line 113: For any lambda that does not pass up a specific screen event, I'd use more descriptive names of what they actually handle. In this case, something like `onToggleVisibility: (Boolean) -> Unit` is way more readable

app/src/main/java/com/example/taskyy/ui/viewmodels/AgendaViewModel.kt
```diff
@@ -62,9 +65,19 @@ class AgendaViewModel @Inject constructor(
...
+            }
+
+            is AgendaEvent.TaskItemSelected -> {
+                savedStateHandle["state"] = state
+
+            }
+
+            is AgendaEvent.EventItemSelected -> {
+                savedStateHandle["state"] = state
+            }
```
philipplackner, line 80: Not really sure what you're trying to do with the saved state handle here

app/src/main/java/com/example/taskyy/ui/viewmodels/ReminderViewModel.kt
```diff
@@ -1,33 +1,124 @@
...
+import java.time.LocalDateTime
+import java.time.ZoneId
+import java.time.format.DateTimeFormatter
+import java.util.Locale
 import javax.inject.Inject
 
 @HiltViewModel
 class ReminderViewModel @Inject constructor(
-    private val sharedPreferences: SharedPreferences
+    private val sharedPreferences: SharedPreferences,
```
philipplackner, line 22: Here, you're coupling a data related dependency to the presentation layer.

The following allows you to easily access user data from any layer in your app while being able to also easily swap out preferences implementations. The only thing needed for that would be to swap out the shared preferences object in the `UserPreferencesImpl` constructor with let's say a DataStore instance.

*Domain model*
```
data class AuthenticatedUser(
    val fullName: String,
    val email: String,
    val token: String
)
```

*Domain abstraction*
```
interface UserPreferences {
    fun saveAuthenticatedUser(user: AuthenticatedUser)
    fun getAuthenticatedUser(): AuthenticatedUser?
    fun clearPreferences()
}
```

*Data implementation*
```
class UserPreferencesImpl(
    private val sharedPref: SharedPreferences
): UserPreferences {
    override fun saveAuthenticatedUser(user: AuthenticatedUser) {
        // Save JSON of user in shared pref
    }

    override fun getAuthenticatedUser(): AuthenticatedUser? { ... }

    override fun clearPreferences() { ... }
}
```

In your Dagger module:
```
@Provides
@Singleton
fun provideUserPreferences(sharedPreferences: SharedPreferences): UserPreferences {
    return UserPreferencesImpl(sharedPreferences)
}
```

Then, for example in any ViewModel:
```
@HiltViewModel
class MyViewModel @Inject constructor(
    private val userPreferences: UserPreferences
): ViewModel() {
    // Use userPreferences
}
```

philipplackner (approved):

PR #23, opened 2024-03-25
Started adding functionality to the reminder screen but realized that…

Mooky4591:
…

### DIFF
--- a/app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
+++ b/app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
@@ -15,9 +15,11 @@ import androidx.navigation.navigation
 import com.example.taskyy.ui.events.AgendaEvent
 import com.example.taskyy.ui.events.LoginEvent
 import com.example.taskyy.ui.events.RegisterEvent
+import com.example.taskyy.ui.events.ReminderEvent
 import com.example.taskyy.ui.screens.AgendaScreen
 import com.example.taskyy.ui.screens.LoginScreen
 import com.example.taskyy.ui.screens.RegisterScreen
+import com.example.taskyy.ui.screens.ReminderDescription
 import com.example.taskyy.ui.screens.ReminderScreen
 import com.example.taskyy.ui.viewmodels.AgendaViewModel
 import com.example.taskyy.ui.viewmodels.LoginViewModel
@@ -90,18 +92,11 @@ fun Nav() {
                 val agendaViewModel = hiltViewModel<AgendaViewModel>()
                 val state = agendaViewModel.state
 
-                ObserveAsEvents(agendaViewModel.events) { event ->
-                    when (event) {
-                        is AgendaEvent.LogoutSuccessful -> navController.navigate(Screen.Login.route)
-                        else -> {}
-                    }
-
-                }
                 AgendaScreen(
                     state = state,
                     onEvent = { event ->
                         when (event) {
-                            is AgendaEvent.ReminderItemSelected -> navController.navigate(Screen.Reminder.route)
+                            is AgendaEvent.ReminderItemSelected -> navController.navigate(Screen.Reminder.route + "/${state.dateString}")
                             else -> agendaViewModel.onEvent(event)
                         }
                         agendaViewModel.onEvent(event = event)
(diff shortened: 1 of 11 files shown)


app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
```diff
@@ -90,32 +92,43 @@ fun Nav() {
...
+
+            composable(route = Screen.ReminderDetails.route) {
+                val reminderViewModel = hiltViewModel<ReminderViewModel>()
+                val state by reminderViewModel.state.collectAsState()
+                ReminderDescription(
+                    state = state,
+                    onEvent = { event ->
+                        reminderViewModel.onEvent(event)
                     }
                 )
```
philipplackner, line 133: As you were already curious about, I'd recommend keeping responsibilities separate. That means you almost always want to have one ViewModel per screen. The only exception is if you need to share state between multiple screens to put that shared state and its logic into a separate shared ViewModel. But you don't need that too often. 

Here, I'd go with a separate `EditTextViewModel` or similar. You will need this same screen also for events and tasks, so you can just reuse it without coupling it to reminders.

app/src/main/java/com/example/taskyy/domain/navigation/Screen.kt
```diff
@@ -4,7 +4,10 @@ sealed class Screen(val route: String) {
     data object Login : Screen(route = "login")
     data object Register : Screen(route = "register")
     data object Agenda : Screen(route = "agenda")
+
+    //reminder screens
     data object Reminder : Screen(route = "reminder")
+    data object ReminderDetails : Screen(route = "reminder_details")
     data object Task : Screen(route = "task")
     data object Event : Screen(route = "event")
```
philipplackner, line 12: It could also make sense to think about if you could have one agenda details screen, since there's a lot of shared logic for each item (e.g. editing title/description, setting the start time, setting the reminder type)

app/src/main/java/com/example/taskyy/ui/ReminderType.kt
```diff
@@ -0,0 +1,8 @@
+package com.example.taskyy.ui
+
+enum class ReminderType {
+    TEN_MINUTES_BEFORE,
+    FIFTEEN_MINUTES_BEFORE,
```
philipplackner, line 5: I think there's no option for 15 minutes on the mockups, is it? On the other hand, you're missing 12h and one day

app/src/main/java/com/example/taskyy/ui/viewmodels/ReminderViewModel.kt
```diff
@@ -69,6 +79,35 @@ class ReminderViewModel @Inject constructor(
...
+        if (timeString == "12:0") {
+            timeString = ("12:00")
+        }
+        val twelveHourDateFormat = DateTimeFormatter.ofPattern("HH:mm", Locale.ENGLISH)
+        val time24 = LocalTime.parse(timeString, twelveHourDateFormat)
+        val formatter12 = DateTimeFormatter.ofPattern("hh:mm a")
+        val time12 = time24.format(formatter12)
+
+        _state.update { it.copy(selectedTime = time12) }
+    }
```
philipplackner, line 96: This function I'm a bit confused about.. Why do you pass the time to format as a String? Normally, if you want to format times, they should be passed in a recognized time format. That is either a Long timestamp, a UTC String timestamp (like `2016-06-01T14:46:22.001Z`) or ideally in the presentation layer an already parsed `LocalTime` instance which you can use to format the string directly.

app/src/main/java/com/example/taskyy/ui/viewmodels/ReminderViewModel.kt
```diff
@@ -69,6 +79,35 @@ class ReminderViewModel @Inject constructor(
...
+
+    private fun setReminderText(alarmTimeText: ReminderType): String {
+        return when (alarmTimeText) {
+            ReminderType.ONE_HOUR_BEFORE -> "1 hour before"
+            ReminderType.THIRTY_MINUTES_BEFORE -> "30 minutes before"
+            ReminderType.FIFTEEN_MINUTES_BEFORE -> "15 minutes before"
+            ReminderType.TEN_MINUTES_BEFORE -> "10 minutes before"
+        }
+    }
```
philipplackner, line 109: This mapping logic is also needed for the other agenda item types, could it make sense to make it a global extension function? (`fun ReminderType.toReminderText(): String`)

philipplackner (approved):

app/src/main/java/com/example/taskyy/ui/viewmodels/ReminderViewModel.kt
Mooky4591: The time picker was giving me integers so I went with it.

PR #24, opened 2024-03-28
I am trying to make the EditTextScreen reusable but I am struggling w…

Mooky4591:
…ith getting that to happen. I am trying to use an Enum of type EditTextScreenType to determine what gets updated and displayed but I can't get the values from the Reminder screen to the Edit Text screen. My screen type is always null

### DIFF
--- a/app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
+++ b/app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
@@ -4,24 +4,29 @@ import android.os.Build
 import android.widget.Toast
 import androidx.annotation.RequiresApi
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.collectAsState
 import androidx.compose.runtime.getValue
 import androidx.compose.ui.platform.LocalContext
 import androidx.hilt.navigation.compose.hiltViewModel
+import androidx.lifecycle.compose.collectAsStateWithLifecycle
 import androidx.navigation.compose.NavHost
 import androidx.navigation.compose.composable
 import androidx.navigation.compose.rememberNavController
 import androidx.navigation.navigation
+import com.example.taskyy.ui.enums.EditTextScreenType
 import com.example.taskyy.ui.events.AgendaEvent
+import com.example.taskyy.ui.events.EditTextEvent
 import com.example.taskyy.ui.events.LoginEvent
 import com.example.taskyy.ui.events.RegisterEvent
 import com.example.taskyy.ui.events.ReminderEvent
 import com.example.taskyy.ui.screens.AgendaScreen
+import com.example.taskyy.ui.screens.EditTextScreen
 import com.example.taskyy.ui.screens.LoginScreen
 import com.example.taskyy.ui.screens.RegisterScreen
-import com.example.taskyy.ui.screens.ReminderDescription
 import com.example.taskyy.ui.screens.ReminderScreen
 import com.example.taskyy.ui.viewmodels.AgendaViewModel
+import com.example.taskyy.ui.viewmodels.EditTextViewModel
 import com.example.taskyy.ui.viewmodels.LoginViewModel
 import com.example.taskyy.ui.viewmodels.RegisterViewModel
 import com.example.taskyy.ui.viewmodels.ReminderViewModel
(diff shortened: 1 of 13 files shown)


app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
```diff
@@ -107,29 +112,82 @@ fun Nav() {
...
+                    .currentBackStackEntry
+                    ?.savedStateHandle
+                    ?.getStateFlow<String>("reminderDescription", "Reminder Description")
+                    ?.collectAsStateWithLifecycle()
+
+                LaunchedEffect(editedRemindDescription) {
+                    reminderViewModel.setReminderDescription(
+                        editedRemindDescription?.value ?: "Reminder Description"
+                    )
+                }
```
philipplackner, line 125: Ah, you might need to refer to `editedRemindDescription.value` here for the key of LaunchedEffect as well, otherwise Compose will listen to changes of the whole state field and not its value.
Alternatively, you can use `by` to avoid having to refer to `.value` all the time: 
```
val editedReminderDescription by navController.currentBackStackEntry...
```

app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
```diff
@@ -107,29 +112,82 @@ fun Nav() {
...
-                            is ReminderEvent.EnterReminderDescription -> navController.navigate(
-                                Screen.ReminderDetails.route
+                    state = state
+                ) { event ->
+                    when (event) {
+                        is ReminderEvent.EnterReminderDescription -> {
+                            navController.previousBackStackEntry?.savedStateHandle?.set(
+                                "screenType",
+                                event.editDescription
                             )
```
philipplackner, line 135: You're here on the reminder detail screen and edit the previous back stack entry, so the one of the agenda screen. That's not correct

app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
```diff
@@ -107,29 +112,82 @@ fun Nav() {
...
+                            navController.navigate(Screen.EditText.route)
```
philipplackner, line 136: Navigating to a screen with parameter (which type of edit text field it is) still works with normal Compose navigation arguments, but here you don't pass any to that navigation. The route should look something like this:
```
"edit_text/{screenType}
```
And then navigating to it like this:
```
navController.navigate("edit_text/${ScreenType.TITLE}")
```

app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
```diff
@@ -107,29 +112,82 @@ fun Nav() {
...
+                                )
+                                navController.popBackStack()
+                            }
+
+                            is EditTextEvent.SaveTitle -> {
+                                navController.previousBackStackEntry?.savedStateHandle?.set(
+                                    "reminderTitle",
+                                    state.enteredText
+                                )
+                            }
```
philipplackner, line 187: This looks good, you're just missing the `popBackStack` after save title

app/src/main/java/com/example/taskyy/ui/enums/EditTextScreenType.kt
```diff
@@ -0,0 +1,7 @@
+package com.example.taskyy.ui.enums
+
+enum class EditTextScreenType {
+    EDIT_DETAILS,
+    EDIT_TITLE,
+    NULL
```
philipplackner, line 6: This null value doesn't really make sense. Kotlin lets you model anything as a null value, so you don't need your own ones. The enum is fine with just the first two options. To indicate there is no value, you can still do 
```
val myEnum: EditTextScreenType? = null
```

app/src/main/java/com/example/taskyy/ui/viewmodels/AgendaViewModel.kt
```diff
@@ -124,23 +129,31 @@ class AgendaViewModel @Inject constructor(
...
+            state = state.copy(initials = savedStateHandle.get<String>("userInitials") ?: "")
+        } else {
+            viewModelScope.launch {
+                val name = agendaRepository.getUserName(email!!)
+                sharedPreferences.edit().putString("name", name).apply()
+                state = state.copy(name = name)
+                state = state.copy(initials = name
+                    .split(' ')
+                    .mapNotNull { it.firstOrNull()?.toString() }
+                    .reduce { acc, s -> acc + s })
```
philipplackner, line 142: If the user has a middle name, this will result in 3 initials being shown. In that case I'd expect that only the first and last name's initials will be shown

philipplackner (approved):

PR #25, opened 2024-03-29
Removed some unnecessary code, got the description and title fields w…

Mooky4591:
…orking. For some reason the time and data picker broke and the state doesn't update when I select new time and date values from the pickers

### DIFF
--- a/app/src/main/java/com/example/taskyy/ui/screens/ReminderScreen.kt
+++ b/app/src/main/java/com/example/taskyy/ui/screens/ReminderScreen.kt
@@ -55,7 +55,6 @@ import com.example.taskyy.ui.viewmodels.ReminderState
 @Composable
 fun ReminderScreen(state: ReminderState, onEvent: (ReminderEvent) -> Unit) {
     Scaffold {
-        onEvent(ReminderEvent.SetUserDefaults)
         Column(
             verticalArrangement = Arrangement.SpaceEvenly,
             modifier = Modifier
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/example/taskyy/ui/screens/ReminderScreen.kt
```diff
@@ -431,7 +436,9 @@ fun TimePickerDialog(
                     ) { Text("Cancel") }
                     TextButton(
                         onClick = {
-                            timeSelected("${pickedTime.hour}:${pickedTime.minute}")
+                            val hour = pickedTime.hour
+                            val minute = pickedTime.minute
```
philipplackner, line 440: Why not just pass `pickedTime` here which includes all the information about the time you need? Formatting it as a `$hour:$minute` string just makes extra effort having to convert it back to time values

app/src/main/java/com/example/taskyy/ui/screens/ReminderScreen.kt
```diff
@@ -305,40 +304,46 @@ fun DateSection(
...
         TimeSection(
             isTimePickerExpanded = isTimePickerExpanded,
-            onEvent = onTimeSelected,
-            selectedTime = selectedTime
+            selectedTime = timeString,
+            toggleTimePicker = {
+                onTimeExpanded(it)
+            },
+            userSelectedTime = {
+                ReminderEvent.TimeSelected(selectedTime = it)
```
philipplackner, line 324: Here you're not responding to the lambda, you just create the event instance which doesn't do anything at this point. That's why your state isn't updating

philipplackner (approved):

PR #26, opened 2024-03-30
Finished the save functionality of the reminder

Mooky4591:


app/build.gradle
```diff
@@ -127,6 +127,10 @@ dependencies {
     implementation 'androidx.compose.ui:ui:1.6.4'
     implementation 'androidx.compose.material:material:1.6.4'
 
+    // Local Time
+    implementation 'joda-time:joda-time:2.10.12'
```
philipplackner, line 131: This is quite an old dependency, you're very fine with the normal Java LocalDate time. How comes you felt like you needed this?

app/src/main/java/com/example/taskyy/data/repositories/AgendaRepositoryImpl.kt
```diff
@@ -30,4 +36,43 @@ class AgendaRepositoryImpl @Inject constructor(
         return userDao.getUserNameByEmail(email)
     }
 
+    override suspend fun saveReminder(reminder: Reminder): com.example.taskyy.domain.error.Result<Reminder, DataError.Network> {
+        return try {
+            val reminderEntity = reminder.toReminderEntity()
+            val reminderDTO = reminder.toReminderDto()
+            retrofit.createReminder(reminderDTO)
+            agendaDao.insertReminder(reminderEntity)
```
philipplackner, line 44: Tasky should be an offline-first app, so fully usable in offline mode. When saving such items, you need to take a bit of care with the order you insert them. If the Retrofit call fails, the item will never be inserted into the local DB here, even though that's what someone would expect from an offline-first app.
The general flow would need to be this:
1. Insert item in DB
2. Post item to remote API. Success? Finish here.
3. Posting to API failed? Keep a reference to this item, so you can try to sync it later (when internet is back or the call doesn't fail for whatever reason)

app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
```diff
@@ -135,6 +136,22 @@ fun Nav() {
                     )
                 }
 
+                ObserveAsEvents(reminderViewModel.events) { event ->
+                    when (event) {
+                        is ReminderEvent.SaveFailed -> Toast.makeText(
+                            context,
+                            event.errorMessage.asString(context),
+                            Toast.LENGTH_SHORT
+                        ).show()
```
philipplackner, line 145: Therefore, you usually also wouldn't tell the user that saving failed, since saving locally will always be successful (unless the device storage is full). You could tell them that the app couldn't sync the note, but not that it generally failed to save it

app/src/main/java/com/example/taskyy/ui/viewmodels/ReminderViewModel.kt
```diff
@@ -77,34 +99,106 @@ class ReminderViewModel @Inject constructor(
...
+        _state.update { it.copy(selectedTime = formattedTime) }
+    }
+
+    private fun save(title: String, description: String) {
+        val newReminder = Reminder(
+            alarmType = _state.value.formattedReminderType,
+            title = title,
+            description = description,
+            timeInMillis = _state.value.formattedReminderTime,
+            id = Random.nextInt(1..10000),
```
philipplackner, line 119: The IDs for items are strings in this app. It's quite important that the client generates a unique string ID, so it also works with the backend, even when it gets synced later

app/src/main/java/com/example/taskyy/ui/viewmodels/ReminderViewModel.kt
```diff
@@ -77,34 +99,106 @@ class ReminderViewModel @Inject constructor(
...
-    private fun save() {
-        TODO("Not yet implemented")
+    private fun formatReminderType(reminderType: String, timeToAdd: Int) {
+        val timeToAddInMilis: Long = when (reminderType) {
+            "10 minutes before" -> timeToAdd * 60 * 1000L
+            "30 minutes before" -> timeToAdd * 60 * 1000L
+            "1 hour before" -> timeToAdd * 60 * 60 * 1000L
+            "6 hours before" -> timeToAdd * 60 * 60 * 1000L
+            "1 day before" -> timeToAdd * 24 * 60 * 60 * 1000L
+            else -> {
```
philipplackner, line 157: For values like `timeToAdd` I'd always also specify the unit you expect, that's otherwise not clear without trying to understand the code. Here you're talking about minutes I think? So it could make sense calling it `minutesToAdd`

app/src/main/java/com/example/taskyy/ui/viewmodels/ReminderViewModel.kt
```diff
@@ -77,34 +99,106 @@ class ReminderViewModel @Inject constructor(
...
+                }
+            }
         }
-        val twelveHourDateFormat = DateTimeFormatter.ofPattern("HH:mm", Locale.ENGLISH)
-        val time24 = LocalTime.parse(timeString, twelveHourDateFormat)
-        val formatter12 = DateTimeFormatter.ofPattern("hh:mm a")
-        val time12 = time24.format(formatter12)
+    }
+
+    private fun formatTimeForDB(selectedTime: String, selectedDate: String) {
```
philipplackner, line 135: I wouldn't create my own date format for the DB. There are standardized formats every date formatter can understand. That's typically either storing the Long timestamp or an ISO date string (like `2011-10-05T14:48:00.000Z`)

philipplackner (approved):

app/build.gradle
Mooky4591: I was trying to get a function to work and This made it work. I will try and remove it and see if it still works.

app/src/main/java/com/example/taskyy/ui/viewmodels/ReminderViewModel.kt
Mooky4591: So how do I get this value? Nothing gets returned so should the user generate this ID? Also, do you have any recommendations for keeping a reference to this reminder if it doesn't save remotely and then upload it when Internet connection is established? This isn't something I've dealt with before so I'm unsure of the best way to do that

app/src/main/java/com/example/taskyy/ui/viewmodels/ReminderViewModel.kt
Mooky4591: Well it's whatever value the user has selected. If the user selected 10 minutes before then timeToAdd is minutes but if they selected 1 hour before then the timeToAdd value is hours. Originally I had an Enum value passed in to determine if it was hours or minutes but I re worked it to this because the code seemed simpler to me

app/src/main/java/com/example/taskyy/ui/viewmodels/ReminderViewModel.kt
Mooky4591: Well the date and time are seperate strings so I have to convert them both to milliseconds separately and then add them together to get the date and time in milliseconds. I'm not sure how else I would do it

app/build.gradle
Mooky4591: Yeah this portion doesn't work without it.
Screenshot 2024-03-30 at 3 05 17 PM

philipplackner: And what's the error? Related to API level 26?

app/src/main/java/com/example/taskyy/ui/viewmodels/ReminderViewModel.kt
philipplackner: You can generate IDs with `UUID.randomUUID().toString()`

There are many ways to implement a syncing strategy. I personally added a table to the DB where I save which agenda item was locally modified and in which way. E.g. event ID abc was locally created.
Then a WorkManager worker can regularly go through this DB table and sync the pending items and remove them afterwards

app/src/main/java/com/example/taskyy/ui/viewmodels/ReminderViewModel.kt
philipplackner: Hm, that doesn't feel very consistent to me and wasn't clear at all. Why to you need this time to add at all? You could also perfectly work with the Kotlin Duration class:
```
"10 minutes before" -> 10.minutes.toLong(DurationUnit.MILLISECONDS)
...
"6 hours before" -> 6.hours.toLong(DurationUnit.MILLISECONDS)
```

app/src/main/java/com/example/taskyy/ui/viewmodels/ReminderViewModel.kt
philipplackner: You can work with one single `LocalDateTime` state which contains both date and time information. You can then format this instance once as HH:mm and once for the date

Mooky4591: How would that work? I think I understand what your saying but I can't visualize what the implementation would look like

app/build.gradle
Mooky4591: Screenshot 2024-03-30 at 10 26 49 PM

philipplackner: Ah, I see. It just expects a different way to create it: `LocalTime.of(time.hour, time.minute)`

app/src/main/java/com/example/taskyy/ui/viewmodels/ReminderViewModel.kt
philipplackner: Something like this:
```
data class MyScreenState(
    val dateTime: LocalDateTime = LocalDateTime.now(),
    ...
)
```
And in your UI:
```
val formattedDate = remember(state.dateTime) {
    DateTimeFormatter.ofPattern("dd MM yyyy").format(state.dateTime)
}
val formattedTime = remember(state.dateTime) {
    DateTimeFormatter.ofPattern("HH mm").format(state.dateTime)
}
```

philipplackner: And when you select a `LocalDate` from the UI and want to set that in your LocalDateTime instance of the state, you should be able to do this:
```
state = state.copy(dateTime = state.dateTime.with(newChosenDate)
```

PR #27, opened 2024-04-01
I am working to refactor the way I'm doing the date/time. Once I get …

Mooky4591:
…this resolved, the newly created reminders should show on the agenda screen. However, during the refactor, I can no longer get the sharedPref to work. Specifically, I can see that I am adding the email on a successful login but when I try to get the email back out of sharedPref, it is empty and I'm not sure why


app/src/main/java/com/example/taskyy/data/repositories/UserPreferencesImpl.kt
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    override fun addUserId(userId: String, key: String) {
+        sharedPref.edit().putString(userId, key).apply()
+    }
+
+    override fun getUserId(key: String): String {
+        val userId = sharedPref.getString(key, "")
+        return userId!!
+    }
```
philipplackner: The function signature of getUserEmail and getUserId is identical. I'd argue it's not the responsibility of the caller to know how preferences are working internally (and that it works with keys and it has to know which key results in which piece of data). Better keep the keys private to this class and just pass the right key for the right function

app/src/main/java/com/example/taskyy/data/repositories/UserPreferencesImpl.kt
```diff
@@ -0,0 +1,57 @@
...
+        sharedPref.edit().putString(fullName, key).apply()
+    }
+
+    override fun getUserFullName(key: String): String {
+        val name = sharedPref.getString(key, "")
+        return name!!
+    }
+
+    override fun addUserEmail(email: String, key: String) {
+        sharedPref.edit().putString(email, key).apply()
```
philipplackner: I think the key comes first and then the email. That's likely why you're not getting it back

philipplackner (approved):

app/src/main/java/com/example/taskyy/data/local/room_database/TaskyyDatabase.kt
```diff
@@ -6,10 +6,10 @@ import androidx.room.Room
 import androidx.room.RoomDatabase
 import com.example.taskyy.data.local.data_access_objects.AgendaActivityDao
 import com.example.taskyy.data.local.data_access_objects.UserDao
-import com.example.taskyy.data.local.room_entity.ReminderEntity
-import com.example.taskyy.data.local.room_entity.UserEntity
+import com.example.taskyy.data.local.room_entity.agenda_entities.ReminderEntity
+import com.example.taskyy.data.local.room_entity.user.UserEntity
 
-@Database(entities = [UserEntity::class, ReminderEntity::class], version = 4, exportSchema = false)
+@Database(entities = [UserEntity::class, ReminderEntity::class], version = 5, exportSchema = false)
```
philipplackner: Increasing the version is meant for migrations, so if your app is installed on some user's devices and then there's an update available with a different DB schema. The version is then used to detect changes in the DB schema to properly migrate.

If you just locally build an app like here, this can always stay at 1. If you change your schema and get a crash, you just need to reinstall the app (completely uninstall on your device)

app/src/main/java/com/example/taskyy/data/repositories/AgendaRepositoryImpl.kt
```diff
@@ -36,31 +39,78 @@ class AgendaRepositoryImpl @Inject constructor(
...
+            Result.Success(reminder)
+        } catch (e: IOException) {
+            when (e.message) {
+                "Permission denied" -> Result.Error(DataError.Local.PERMISSION_DENIED)
+                "File not found" -> Result.Error(DataError.Local.FILE_NOT_FOUND)
+                "Disk full" -> Result.Error(DataError.Local.DISK_FULL)
+                "Input/output error" -> Result.Error(DataError.Local.INPUT_OUTPUT_ERROR)
+                "Connection refused" -> Result.Error(DataError.Local.CONNECTION_REFUSED)
+                else -> Result.Error(DataError.Local.UNKNOWN)
+            }
```
philipplackner, line 55: Are these errors here even relevant? When would you get file not found? Or permission denied? The only error I can think of is `SQLiteFullException` when the disk is full

app/src/main/java/com/example/taskyy/data/repositories/AgendaRepositoryImpl.kt
```diff
@@ -36,31 +39,78 @@ class AgendaRepositoryImpl @Inject constructor(
...
-                408 -> com.example.taskyy.domain.error.Result.Error(DataError.Network.REQUEST_TIMEOUT)
-                429 -> com.example.taskyy.domain.error.Result.Error(DataError.Network.TOO_MANY_REQUESTS)
-                413 -> com.example.taskyy.domain.error.Result.Error(DataError.Network.PAYLOAD_TOO_LARGE)
-                500 -> com.example.taskyy.domain.error.Result.Error(DataError.Network.SERVER_ERROR)
-                400 -> com.example.taskyy.domain.error.Result.Error(DataError.Network.SERIALIZATION)
-                else -> com.example.taskyy.domain.error.Result.Error(DataError.Network.UNKNOWN)
+                408 -> Result.Error(DataError.Network.REQUEST_TIMEOUT)
+                429 -> Result.Error(DataError.Network.TOO_MANY_REQUESTS)
+                413 -> Result.Error(DataError.Network.PAYLOAD_TOO_LARGE)
+                500 -> Result.Error(DataError.Network.SERVER_ERROR)
```
philipplackner, line 69: Any code between 500 and 599 could be a server error, so you could catch them all with `in 500..599 ->`

app/src/main/java/com/example/taskyy/data/repositories/AgendaRepositoryImpl.kt
```diff
@@ -36,31 +39,78 @@ class AgendaRepositoryImpl @Inject constructor(
...
+        Reminder(
+            reminderEntity.title,
+            reminderEntity.description,
+            reminderEntity.time,
+            reminderEntity.remindAt,
+            reminderEntity.id,
+            reminderEntity.userId!!
+        )
+    }
 }
```
philipplackner, line 107: If you make this function extend just `ReminderEntity` it's more flexible since it also works with individual reminders outside of a list. For a list you can still do `map { it.toReminder() }`

philipplackner (approved):

PR #28, opened 2024-04-08
Made some changes and refactored. I am trying to get the list of agen…

Mooky4591:
…da events to be functional. I left some notes in the code where I had questions but I am trying to pass an enum through the navController, into the savedStateHandle but it is landing in the reminder view model as a string. I had to create an extension function to convert it back. That doesn't feel very efficient. Also I'm having a timing issue where I am grabbing the reminder from the DB and setting the values when opening up an already existing reminder. However, the values get set after already landing on the reminder page.

### DIFF
--- a/app/src/main/java/com/example/taskyy/data/local/room_entity/agenda_entities/ReminderEntity.kt
+++ b/app/src/main/java/com/example/taskyy/data/local/room_entity/agenda_entities/ReminderEntity.kt
@@ -7,9 +7,9 @@ import androidx.room.PrimaryKey
 data class ReminderEntity(
     @PrimaryKey
     val id: String,
-    val userId: String?,
     val description: String,
     val remindAt: Long,
     val time: Long,
-    val title: String
+    val title: String,
+    val color: String
 )
--- a/app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
+++ b/app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
@@ -16,6 +16,7 @@ import androidx.navigation.compose.composable
 import androidx.navigation.compose.rememberNavController
 import androidx.navigation.navArgument
 import androidx.navigation.navigation
+import com.example.taskyy.ui.enums.AgendaItemType
 import com.example.taskyy.ui.enums.EditTextScreenType
 import com.example.taskyy.ui.events.AgendaEvent
 import com.example.taskyy.ui.events.EditTextEvent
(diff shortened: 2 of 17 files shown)


app/src/main/java/com/example/taskyy/data/local/room_entity/agenda_entities/ReminderEntity.kt
```diff
@@ -7,9 +7,9 @@ import androidx.room.PrimaryKey
 data class ReminderEntity(
     @PrimaryKey
     val id: String,
-    val userId: String?,
     val description: String,
     val remindAt: Long,
     val time: Long,
-    val title: String
+    val title: String,
+    val color: String
```
philipplackner, line 14: Saving the color is dangerous. You know which agenda item type has which color, so why the need to save it? (and it's not like each item has it's own custom color which is unpredictable) 

Imagine, you want to change the colors in a future update of the app, then you'll quickly mix them with older colors if you don't have a proper DB migration

app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
```diff
@@ -170,6 +176,10 @@ fun Nav() {
                             navController.navigate(Screen.EditText.route + "/${event.editTitle}")
                         }
 
+                        is ReminderEvent.Close -> {
+                            navController.popBackStack()
+                        }
```
philipplackner, line 181: For user-initiated back stack popping, I'd prefer `navigateUp`, since that is aware of your back stack and will only navigate to the previous screen (and not generally just pop one from the back stack whatever it is).
When this is done by `popBackStack`, clicking a button and there's a transition animation or so, it can happen that the user could press the button multiple times when tapped fast. That would clear more of the back stack as you'd want here

app/src/main/java/com/example/taskyy/ui/viewmodels/ReminderViewModel.kt
```diff
@@ -295,6 +333,25 @@ class ReminderViewModel @Inject constructor(
...
+            AgendaItemType.TASK_ITEM
+        }
+
+        AgendaItemType.EVENT_ITEM.toString() -> {
+            AgendaItemType.EVENT_ITEM
+        }
+
+        else -> {
+            AgendaItemType.REMINDER_ITEM
+        }
```
philipplackner, line 352: You can just do `AgendaItemType.valueOf(yourString)`

philipplackner (approved):

philipplackner:
It's very normal that enums are serialized as strings, that's not a problem

PR #29, opened 2024-04-11
Added the functionality for the 3 dot menu on the agenda list items. …

Mooky4591:
…Created the delete and update database functionality

### DIFF
--- a/app/src/main/java/com/example/taskyy/data/local/data_access_objects/AgendaActivityDao.kt
+++ b/app/src/main/java/com/example/taskyy/data/local/data_access_objects/AgendaActivityDao.kt
@@ -1,6 +1,7 @@
 package com.example.taskyy.data.local.data_access_objects
 
 import androidx.room.Dao
+import androidx.room.Delete
 import androidx.room.Upsert
 import com.example.taskyy.data.local.room_entity.agenda_entities.ReminderEntity
 
@@ -15,5 +16,17 @@ interface AgendaActivityDao {
 
     @androidx.room.Query("SELECT * FROM reminder_table WHERE id = :eventID")
     suspend fun getReminderByEventId(eventID: String): ReminderEntity
+
+    @androidx.room.Query("UPDATE reminder_table SET description = :description, time = :time, remindAt = :remindTime, title = :title WHERE id = :eventId")
+    suspend fun updateReminder(
+        description: String,
+        time: Long,
+        title: String,
+        remindTime: Long,
+        eventId: String
+    )
+
+    @Delete
+    suspend fun deleteReminder(reminderEntity: ReminderEntity)
 
 }
--- a/app/src/main/java/com/example/taskyy/data/repositories/AgendaRepositoryImpl.kt
+++ b/app/src/main/java/com/example/taskyy/data/repositories/AgendaRepositoryImpl.kt
@@ -11,6 +11,7 @@ import com.example.taskyy.domain.error.Result
 import com.example.taskyy.domain.repository.AgendaRepository
 import com.example.taskyy.domain.repository.UserPreferences
 import com.example.taskyy.ui.enums.AgendaItemType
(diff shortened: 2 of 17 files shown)


app/src/main/java/com/example/taskyy/data/local/data_access_objects/AgendaActivityDao.kt
```diff
@@ -16,4 +17,16 @@ interface AgendaActivityDao {
...
     suspend fun getReminderByEventId(eventID: String): ReminderEntity
 
+    @androidx.room.Query("UPDATE reminder_table SET description = :description, time = :time, remindAt = :remindTime, title = :title WHERE id = :eventId")
+    suspend fun updateReminder(
+        description: String,
+        time: Long,
+        title: String,
+        remindTime: Long,
+        eventId: String
+    )
```
philipplackner, line 27: Two things:
1. Why don't you just expose a single `@Upsert` function which lets you insert and update events in one function? I find passing those individual fields rather complicated, since you could just insert a whole entity with the changed data. If you insert that with an ID that exists in the DB and you use `@Upsert`, it will replace the existing entity
2. I think you mean `reminderId` here for the parameter

app/src/main/java/com/example/taskyy/data/repositories/AgendaRepositoryImpl.kt
```diff
@@ -74,6 +97,39 @@ class AgendaRepositoryImpl @Inject constructor(
...
+            Result.Success(reminder)
+        } catch (e: HttpException) {
+            when (e.code()) {
+                408 -> Result.Error(DataError.Network.REQUEST_TIMEOUT)
+                429 -> Result.Error(DataError.Network.TOO_MANY_REQUESTS)
+                413 -> Result.Error(DataError.Network.PAYLOAD_TOO_LARGE)
+                500 -> Result.Error(DataError.Network.SERVER_ERROR)
+                400 -> Result.Error(DataError.Network.SERIALIZATION)
+                else -> Result.Error(DataError.Network.UNKNOWN)
+            }
```
philipplackner, line 113: This response code handling could perfectly be outsourced to a helper function, since it'll be the same for every single HTTP call

app/src/main/java/com/example/taskyy/data/repositories/AgendaRepositoryImpl.kt
```diff
@@ -108,6 +164,32 @@ class AgendaRepositoryImpl @Inject constructor(
...
+            Result.Success(true)
+        } catch (e: IOException) {
+            when (e.message) {
+                "Permission denied" -> Result.Error(DataError.Local.PERMISSION_DENIED)
+                "File not found" -> Result.Error(DataError.Local.FILE_NOT_FOUND)
+                "Disk full" -> Result.Error(DataError.Local.DISK_FULL)
+                "Input/output error" -> Result.Error(DataError.Local.INPUT_OUTPUT_ERROR)
+                "Connection refused" -> Result.Error(DataError.Local.CONNECTION_REFUSED)
+                else -> Result.Error(DataError.Local.UNKNOWN)
+            }
```
philipplackner, line 180: Same for this

app/src/main/java/com/example/taskyy/ui/screens/ReminderScreen.kt
```diff
@@ -709,7 +735,22 @@ fun SaveReminder(
         modifier =
         Modifier
             .clickable {
-                onEvent(ReminderEvent.SaveReminder(title = title, description = description))
+                if (isEventBeingEdited && eventId == "") {
```
philipplackner, line 738: Instead of making non-existent IDs empty strings, I'd make them `null` instead. That's what every developer understands of a non-existent value. While having a non-nullable string, it's not immediately obvious that this could potentially be empty meaning it's non-existent

philipplackner (approved):

PR #30, opened 2024-04-12

Mooky4591:


app/src/main/java/com/example/taskyy/data/local/room_entity/agenda_entities/PendingEventRetryEntity.kt
```diff
@@ -0,0 +1,17 @@
...
+data class PendingEventRetryEntity(
+    @PrimaryKey
+    val id: String,
+    val description: String,
+    val remindAt: Long,
+    val time: Long,
+    val title: String,
+    val action: AgendaItemAction
+
+)
```
philipplackner, line 17: Just saving the id and action should be enough, since with that ID you can retrieve the actual event from the real DB. Saving duplicate data like this is very dangerous, because the real event could've already been updated, but not this entity here

app/src/main/java/com/example/taskyy/data/remote/Workers/AgendaItemWorker.kt
```diff
@@ -0,0 +1,76 @@
...
+    private suspend fun checkReminderRetryTable(): Result {
+        val reminderList = taskyyDatabase.pendingReminderRetryDao().getPendingReminderRetries()
+        if (reminderList.isNotEmpty()) {
+            for (reminder in reminderList) {
+                when (reminder.action) {
+                    AgendaItemAction.CREATE -> {
+                        try {
+                            taskyyApi.createReminder(reminder.toReminderDTO())
+                        } catch (e: HttpException) {
+                            return Result.retry()
```
philipplackner, line 42: Here, I'd have a bit more fine grained response handling. For example, if you get 401, you can be sure that the next call will fail as well, so there's no point in retrying. However, if you get 500 (server error), it could be a temporary thing which could be fixed by just retrying the request. So it really depends on the status code what you want to do there

app/src/main/java/com/example/taskyy/ui/screens/AgendaScreen.kt
```diff
@@ -140,6 +156,31 @@ fun AgendaScreen(state: AgendaState, onEvent: (AgendaEvent) -> Unit, timeDateSta
...
+        ).setConstraints(constraints)
+            .build()
+        val workManager = WorkManager.getInstance(context)
+        workManager.enqueueUniquePeriodicWork(
+            "AgendaItemWorker",
+            ExistingPeriodicWorkPolicy.KEEP,
+            workRequest
+        )
+    }
+}
```
philipplackner, line 182: This does not need to be a composable. The UI should not be responsible for handling background syncs and data related stuff. Instead, you can put this in a class in data and have an abstraction in domain which lets you schedule such a worker

philipplackner (approved):

app/src/main/java/com/example/taskyy/data/remote/Workers/AgendaItemWorker.kt
Mooky4591: So if in the event there is 401 error, would I just return Result.failure() and return Result.retry() if the a 500 error?

philipplackner: Correct, those were just examples though. You'll need to decide this for all common status codes that could be thrown

PR #31, opened 2024-04-13

Mooky4591:


app/src/main/java/com/example/taskyy/data/remote/Workers/AgendaItemWorker.kt
```diff
@@ -37,40 +36,64 @@ class AgendaItemWorker(
                 when (reminder.action) {
                     AgendaItemAction.CREATE -> {
                         try {
-                            taskyyApi.createReminder(reminder.toReminderDTO())
+                            taskyyApi.createReminder(
+                                taskyyDatabase.agendaDao().getReminderByEventId(reminder.id)
+                                    .toReminderDTO()
+                            )
+                            taskyyDatabase.pendingReminderRetryDao().removePendingReminder(reminder)
```
philipplackner, line 43: Even though very unlikely, you'll need to take some care when launching two such suspend functions one after another when either both have to be executed or none. Imagine, you'd finish the first call to `taskyyApi.createReminder` and then before the code reaches the next suspend function, the coroutine is cancelled. This could happen when this suspend fun is launched in `viewModelScope` for example and the user navigates back. In that case the line to update the local DB won't be executed, however the API was already notified about that change. So the app still thinks the reminder is pending and tries to sync it the next time, even though it is already synced (and this could lead to invalid states, since your invalid sync could override the reminder on the API which could've been modified already at this point).

The fix is to make sure that the second suspend function is not cancellable. The best way to do this is with a custom application wide coroutine scope, which lives as long as your app's process:
```
class MyApp: Application() {
    val applicationScope = CoroutineScope(SupervisorJob())
    
    // ...
}
```
You can then provide an instance of this scope in your Dagger module:
```
fun provideApplicationScope(app: Application): CoroutineScope {
    return (app as MyApp).applicationScope
}
```
Then, inject it into the class you need it in:
```
class MyRepository(
    private val applicationScope: CoroutineScope,
    ...
)
```
And make sure the second suspend function is executed in this new scope which won't be cancelled:
```
applicationScope.launch {
    // DB call
}.join()
```
This way, if the coroutine reached this point and wasn't cancelled, we make sure that it also won't be cancelled during this second call. The `join()` makes sure to still suspend the outer suspend function until the DB call finished.

app/src/main/java/com/example/taskyy/data/remote/Workers/AgendaItemWorker.kt
```diff
@@ -37,40 +36,64 @@ class AgendaItemWorker(
...
                     }
                 }
             }
         }
         return Result.success()
     }
+
+    private fun shouldTryAgainOrFail(code: Int): Result {
+        return when (code) {
+            in 500..504 -> Result.retry()
```
philipplackner, line 80: Actually all server side errors could indicate a retry, so `500..599`

app/src/main/java/com/example/taskyy/data/repositories/AgendaRepositoryImpl.kt
```diff
@@ -138,6 +150,27 @@ class AgendaRepositoryImpl @Inject constructor(
...
+            duration = Duration.ofSeconds(15)
+        ).setConstraints(constraints)
+            .build()
+        val workManager = WorkManager.getInstance(context)
+        workManager.enqueueUniquePeriodicWork(
+            "AgendaItemWorker",
+            ExistingPeriodicWorkPolicy.KEEP,
+            workRequest
+        )
+    }
```
philipplackner, line 172: This is already much better as what you had before. Ideally, I'd put this in a separate class in the data layer, since scheduling such workers isn't really the typical job of a repository, but you can also leave this here

app/src/main/java/com/example/taskyy/ui/events/AgendaEvent.kt
```diff
@@ -24,4 +25,5 @@ sealed interface AgendaEvent {
     data class EditExistingReminder(val agendaItem: AgendaEventItem) : AgendaEvent
     data class DeleteExistingReminder(val agendaEventItem: AgendaEventItem) : AgendaEvent
     data class IsEllipsisMenuExpanded(val isEllipsisMenuExpanded: Boolean) : AgendaEvent
+    data class StartWorkManager(val context: Context) : AgendaEvent
```
philipplackner, line 28: Why do you need to trigger scheduling this from the UI? You can do this from the `init` in your ViewModel

app/src/main/java/com/example/taskyy/ui/screens/AgendaScreen.kt
```diff
@@ -156,29 +147,12 @@ fun AgendaScreen(state: AgendaState, onEvent: (AgendaEvent) -> Unit, timeDateSta
...
-            ExistingPeriodicWorkPolicy.KEEP,
-            workRequest
-        )
-    }
+fun StartWorkManager(
+    startWorkManager: (AgendaEvent) -> Unit
+) {
+    val context: Context = LocalContext.current
+    startWorkManager(AgendaEvent.StartWorkManager(context))
 }
```
philipplackner, line 156: This also doesn't really do anything as a separate function. I'd recommend to stick to launching it from the VM's init

app/src/main/java/com/example/taskyy/ui/viewmodels/AgendaViewModel.kt
```diff
@@ -103,6 +104,15 @@ class AgendaViewModel @Inject constructor(
                 deleteReminder(event.agendaEventItem)
             }
             is AgendaEvent.EditExistingReminder -> {}
+            is AgendaEvent.StartWorkManager -> {
+                startWorkManager(event.context)
+            }
+        }
+    }
+
+    private fun startWorkManager(context: Context) {
```
philipplackner, line 113: Also be very careful with using context references in the VM, especially the Activity context. If you store this in a field in the ViewModel (which you don't do here), this quickly causes a memory leak when the activity is destroyed due to a screen rotation for example.

philipplackner (approved):

app/src/main/java/com/example/taskyy/ui/events/AgendaEvent.kt
Mooky4591: I triggered it from the UI so that I had a reference to the context. I need the context in the repository to launch the worker. Is there a better way to get the context to the work manager?

app/src/main/java/com/example/taskyy/data/remote/Workers/AgendaItemWorker.kt
Mooky4591: Does this go in the App class? This one?

@HiltAndroidApp
public class App extends Application {
}

When I tried to put it there it gave me an error so I don't think that's where it is supposed to be

PR #32, opened 2024-04-13
Attempting to implement splash screen and do token authorization chec…

Mooky4591:
Screenshot 2024-04-12 at 11 05 50 PM

…k. I am pretty sure this doesn't work but I'm not totally sure how to test it. Also, after implementing the splash screen my theme for the rest of the app got messed up. I'm not sure how since I'm using the same theme in the postSplashScreen tag as I was in the theme on the Manifest

### DIFF
--- a/app/src/main/java/com/example/taskyy/data/repositories/AuthRepositoryImpl.kt
+++ b/app/src/main/java/com/example/taskyy/data/repositories/AuthRepositoryImpl.kt
@@ -6,11 +6,15 @@ import com.example.taskyy.data.remote.TaskyyApi
 import com.example.taskyy.data.remote.data_transfer_objects.RegisterUserDTO
 import com.example.taskyy.data.remote.response_objects.LoginUserResponse
 import com.example.taskyy.domain.error.DataError
+import com.example.taskyy.domain.error.Result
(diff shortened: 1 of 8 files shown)


app/src/main/java/com/example/taskyy/data/repositories/AuthRepositoryImpl.kt
```diff
@@ -53,32 +56,48 @@ class AuthRepositoryImpl @Inject constructor(
...
+                "UnknownHostException" -> Result.Error(DataError.Network.UNKNOWN_HOST_EXCEPTION)
+                else -> Result.Error(DataError.Network.UNKNOWN)
             }
         }
     }
 
+    override suspend fun validateToken(): Boolean {
+        val call = retrofit.checkTokenIsValid()
+        var returnValue = false
+        call.enqueue(object : Callback<Void> {
```
philipplackner, line 88: If you're using suspend functions, you don't need enqueue. You can make your Retrofit call return the response directly, without wrapping it into a `Call` object.

philipplackner (approved):
To fix the theming issue, you might need to replace the `parent` attribute for your main themes.xml with `android:Theme.Material.Light.NoActionBar`

PR #33, opened 2024-04-15

Mooky4591:


app/src/main/java/com/example/taskyy/ui/viewmodels/LoginViewModel.kt
```diff
@@ -33,8 +33,18 @@ class LoginViewModel @Inject constructor(
...
         viewModelScope.launch {
-            if (authRepository.validateToken()) {
-                eventChannel.send(LoginEvent.LoginSuccess(state.email))
+            when (authRepository.validateToken()) {
+                is Result.Error -> {
+                }
+
+                is Result.Success -> {
+                    if (userPreferences.getUserFullName("name") != "" && userPreferences.getUserEmail(
+                            "email"
```
philipplackner, line 42: I think I commented on this before, but determining which exact string keys to pass here shouldn't be the business of the class that uses these preferences. You can just keep these internally in the `UserPreferencesImpl` and not pass a key at all. In the end, if I want to get an email, why would I need to pass a key that matches the email?

app/src/main/java/com/example/taskyy/ui/viewmodels/ReminderViewModel.kt
```diff
@@ -156,14 +158,16 @@ class ReminderViewModel @Inject constructor(
...
+                            }
+
+                            is Result.Error -> {
+                            }
                         }
-                        is Result.Error -> {
-                            agendaRepository.addFailedReminderToRetry(newReminder)
-                        }
-                    }
+                    }.join()
```
philipplackner, line 170: Implementing the exact caching strategy isn't something for the VM, that's data logic. The repository should just expose a `saveReminder` function which can be called by the VM

philipplackner (approved):

PR #34, opened 2024-04-15
Got the task events working. I am having an issue with the list of ev…

Mooky4591:
…ents. When I click on one I get a Key was already used error.

### DIFF
--- a/app/src/main/java/com/example/taskyy/data/local/data_access_objects/ReminderDao.kt
+++ b/app/src/main/java/com/example/taskyy/data/local/data_access_objects/ReminderDao.kt
@@ -6,13 +6,13 @@ import androidx.room.Upsert
 import com.example.taskyy.data.local.room_entity.agenda_entities.ReminderEntity
 
 @Dao
-interface AgendaActivityDao {
+interface ReminderDao {
 
     @Upsert
     suspend fun insertReminder(reminderEntity: ReminderEntity)
 
     @androidx.room.Query("SELECT * FROM reminder_table WHERE time BETWEEN :startTime AND :endTime")
-    suspend fun getReminders(startTime: Long, endTime: Long): List<ReminderEntity>
+    suspend fun getReminders(startTime: Long, endTime: Long): MutableList<ReminderEntity>
 
     @androidx.room.Query("SELECT * FROM reminder_table WHERE id = :eventID")
     suspend fun getReminderByEventId(eventID: String): ReminderEntity
--- a/app/src/main/java/com/example/taskyy/domain/repository/AgendaRepository.kt
+++ b/app/src/main/java/com/example/taskyy/domain/repository/AgendaRepository.kt
@@ -1,27 +1,47 @@
 package com.example.taskyy.domain.repository
 
-import android.content.Context
 import com.example.taskyy.domain.error.DataError
 import com.example.taskyy.domain.error.Result
 import com.example.taskyy.ui.objects.AgendaEventItem
 import com.example.taskyy.ui.objects.Reminder
+import com.example.taskyy.ui.objects.Task
 
 interface AgendaRepository {
     suspend fun logout(): Boolean
     suspend fun getUserName(email: String): String
+
+    //Reminder
     suspend fun saveReminderToDB(reminder: Reminder): Result<Reminder, DataError.Local>
     suspend fun saveReminderToApi(reminder: Reminder): Result<Reminder, DataError.Network>
     suspend fun updateReminderToApi(reminder: Reminder): Result<Reminder, DataError.Network>
     suspend fun getReminders(
         startDate: Long,
         endDate: Long
-    ): Result<List<Reminder>, DataError.Local>
+    ): Result<MutableList<Reminder>, DataError.Local>
 
     suspend fun getReminderByEventId(eventId: String): Result<Reminder, DataError.Local>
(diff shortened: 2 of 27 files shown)


app/src/main/java/com/example/taskyy/data/local/data_access_objects/ReminderDao.kt
```diff
@@ -6,13 +6,13 @@ import androidx.room.Upsert
...
 @Dao
-interface AgendaActivityDao {
+interface ReminderDao {
 
     @Upsert
     suspend fun insertReminder(reminderEntity: ReminderEntity)
 
     @androidx.room.Query("SELECT * FROM reminder_table WHERE time BETWEEN :startTime AND :endTime")
-    suspend fun getReminders(startTime: Long, endTime: Long): List<ReminderEntity>
+    suspend fun getReminders(startTime: Long, endTime: Long): MutableList<ReminderEntity>
```
philipplackner, line 15: I'm pretty sure you don't want/need to return a mutable list here 😄

app/src/main/java/com/example/taskyy/domain/repository/AgendaRepository.kt
```diff
@@ -1,27 +1,47 @@
...
+
+    //Reminder
     suspend fun saveReminderToDB(reminder: Reminder): Result<Reminder, DataError.Local>
     suspend fun saveReminderToApi(reminder: Reminder): Result<Reminder, DataError.Network>
     suspend fun updateReminderToApi(reminder: Reminder): Result<Reminder, DataError.Network>
     suspend fun getReminders(
         startDate: Long,
         endDate: Long
-    ): Result<List<Reminder>, DataError.Local>
+    ): Result<MutableList<Reminder>, DataError.Local>
```
philipplackner, line 20: Same here, I don't yet see why this would be necessary. If you really need a mutable list somewhere, you can just convert a normal list to a mutable one at the place where you need it, but you don't need to change it all the way back to the origin of the list (your DB)

app/src/main/java/com/example/taskyy/domain/usecases/CheckForTasksUserCase.kt
```diff
@@ -0,0 +1,25 @@
...
+    suspend fun checkForTasks(dateTime: LocalDateTime): Result<List<Task>, DataError.Local> {
+        val startDateTime = dateTime.withHour(0).withMinute(0).withSecond(0).withNano(0)
+        val endDateTime = dateTime.withHour(23).withMinute(59).withSecond(59).withNano(999_999_999)
+
+
+        return agendaRepository.getTasks(
+            startDate = startDateTime.toMillis(),
+            endDate = endDateTime.toMillis()
+        )
+    }
```
philipplackner, line 24: Here it could make sense to return the result as a Flow, so it's reactive and will let the UI know about new tasks whenever the DB gets updated. Right now, you'll just fetch them once with this.

app/src/main/java/com/example/taskyy/ui/objects/Reminder.kt
```diff
@@ -13,4 +16,31 @@ class Reminder(
...
+
+    fun toReminderDto(): ReminderDTO {
+        return ReminderDTO(
+            id = eventId,
+            description = description,
+            title = title,
+            time = timeInMillis,
+            remindAt = alarmType
+        )
+    }
```
philipplackner, line 38: Not the place to put mappers. Here you're coupling the entities to your UI layer apparently (at least if we look at the package name). The agenda item model is a project-wide domain model and the mappers for that is data logic, so it should be in your app's data layer

app/src/main/java/com/example/taskyy/ui/screens/AgendaScreen.kt
```diff
@@ -499,10 +503,18 @@ fun AgendaDisplayItem(
...
+            AgendaItemType.REMINDER_ITEM -> {
+                Color(android.graphics.Color.parseColor("#f2f6ff"))
+            }
+
+            AgendaItemType.TASK_ITEM -> {
+                Color(android.graphics.Color.parseColor("#289c74"))
+            }
+
+            else -> {
+                Color(android.graphics.Color.parseColor("#FFFFFF"))
```
philipplackner, line 505: You don't need to do this with the old Android colors, you can encode the color code as a Hex Long like this: `Color(0xFF289c74)`. The `0x` indicates the number is in hex format, the first `FF` denotes the alpha value and everything that follows after is the color code.

app/src/main/java/com/example/taskyy/ui/viewmodels/ReminderViewModel.kt
```diff
@@ -171,51 +179,114 @@ class ReminderViewModel @Inject constructor(
...
+                        description = description,
+                        timeInMillis = _dateTimeState.value.dateTime.toMillis(),
+                        id = UUID.randomUUID().toString(),
+                        agendaItem = AgendaItemType.REMINDER_ITEM,
+                        agendaAction = AgendaItemAction.CREATE,
+                    )
+                viewModelScope.launch {
+                    when (val save = agendaRepository.saveTaskToDB(newTask)) {
+                        is Result.Success -> {
+                            when (val save = agendaRepository.saveTaskToDB(newTask)) {
```
philipplackner, line 229: Here, of course my previous feedback also applies that this logic goes in the repository. However, I don't understand why you're calling the save to DB function twice in a row. Did you mean to save it to the API instead?

philipplackner (approved):

PR #35, opened 2024-04-17
4 14 2025 notifications

Mooky4591:
Created the notification classes and made suggested changes. The Notification permissions work but the alarm permissions don't and I can't figure out why. They are exact copies so I'm not sure why it doesn't work

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -3,12 +3,8 @@
     xmlns:tools="http://schemas.android.com/tools">
 
     <uses-permission android:name="android.permission.INTERNET" />
-
     <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
-
     <uses-permission android:name="android.permission.SCHEDULE_EXACT_ALARM" />
-
-    <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
 
     <application
         android:name=".App"
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -4,6 +4,12 @@
 
     <uses-permission android:name="android.permission.INTERNET" />
 
+    <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
+
+    <uses-permission android:name="android.permission.SCHEDULE_EXACT_ALARM" />
+
+    <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
+
     <application
         android:name=".App"
         android:allowBackup="false"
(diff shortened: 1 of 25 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -24,7 +26,51 @@
 
                 <category android:name="android.intent.category.LAUNCHER" />
             </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <data
+                    android:scheme="tasky://"
+                    android:host="www.myapp.com"
```
philipplackner, line 34: Why this host? You can just set it to `eventScreen` or so and leave the path prefix

app/src/main/java/com/example/taskyy/data/local/notifications/BootBroadcastReceiver.kt
```diff
@@ -0,0 +1,57 @@
...
+            // Call method to re-schedule your notifications
+            val hasNotificationPermission =
+                if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU) {
+                    ContextCompat.checkSelfPermission(
+                        context,
+                        android.Manifest.permission.POST_NOTIFICATIONS,
+                    ) == PackageManager.PERMISSION_GRANTED
+                } else {
+                    true
+                }
```
philipplackner, line 36: This is not the right place to check that. If you don't have notification permission here, even notifications in a year won't be scheduled. All that counts is if you have this permission right before you want to show a single notification.
Otherwise, if this code runs and doesn't schedule a single notification due to missing permission and the user grants it shortly after, they will still see none of them

app/src/main/java/com/example/taskyy/data/local/notifications/NotificationBroadcastReceiver.kt
```diff
@@ -0,0 +1,44 @@
...
+        if (hasNotificationPermission) {
+            TaskyNotificationService().showNotification(
+                context = context,
+                notificationId = notificationId,
+                title = title,
+                description = description,
+                agendaId = agendaId,
+                agendaType = agendaType,
+            )
+        }
```
philipplackner, line 42: Even here, you don't need to check that. It's just important before you call `notificationManager.notify()`

app/src/main/java/com/example/taskyy/data/local/notifications/NotificationScheduler.kt
```diff
@@ -0,0 +1,51 @@
...
+    fun scheduleNotification(agendaItem: AgendaEventItem) {
+        alarmManager.apply {
+            canScheduleExactAlarms()
+            setExactAndAllowWhileIdle(
+                AlarmManager.RTC_WAKEUP,
+                agendaItem.alarmType,
+                createPendingIntent(agendaItem),
+            )
+        }
+    }
```
philipplackner, line 24: And what do you do below Android S? This function only works above and will crash below

app/src/main/java/com/example/taskyy/data/local/notifications/TaskyNotificationService.kt
```diff
@@ -0,0 +1,75 @@
...
+                .setContentTitle(title)
+                .setContentText(description)
+                .setSmallIcon(R.drawable.taskyy_foreground)
+                .setContentIntent(pendingIntent)
+                .setPriority(NotificationCompat.PRIORITY_DEFAULT)
+                .setAutoCancel(true)
+                .build()
+
+        notificationManager.notify(notificationId, notification)
+    }
```
philipplackner, line 52: This is now where it should be checked

app/src/main/java/com/example/taskyy/ui/screens/ReminderScreen.kt
```diff
@@ -816,3 +828,65 @@ fun SaveTitle(details: String, onEvent: (EditTextEvent) -> Unit) {
...
+            mutableStateOf(
+                ContextCompat.checkSelfPermission(
+                    context,
+                    Manifest.permission.POST_NOTIFICATIONS,
+                ) == PackageManager.PERMISSION_GRANTED,
+            )
+        } else {
+            mutableStateOf(true)
+        }
+    }
```
philipplackner, line 845: If a value never changes in Compose, there's no point of making it Compose state. You can just make it a normal `val`

app/src/main/java/com/example/taskyy/ui/screens/ReminderScreen.kt
```diff
@@ -816,3 +828,65 @@ fun SaveTitle(details: String, onEvent: (EditTextEvent) -> Unit) {
...
+                hasAlarmPermission = isGranted
+                // TODO: Handle permission denial later
+            },
+        )
+    if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.TIRAMISU && !hasAlarmPermission) {
+        SideEffect {
+            alarmPermissionLauncher.launch(Manifest.permission.SCHEDULE_EXACT_ALARM)
+        }
+    }
+}
```
philipplackner, line 892: On Android, there are multiple different protection levels for permissions. They determine how dangerous a permission is and what should be done to grant it. Scheduling exact alarms is not considered as very dangerous that explicitly requires the user's approval, so this requesting is not necessary. That's also why you don't see any dialog for this.
On the other hand, a permission like recording audio is a dangerous one since it could be quickly abused, so the user needs to grant it.

Either way, all permissions an app needs will be listed in the Google Play listing. You can read more about these levels here: https://developer.android.com/guide/topics/manifest/permission-element

philipplackner (approved):

PR #37, opened 2024-04-18
Created event screen. Made some composables more reusable so I could …

Mooky4591:
…use them for all types of agenda items

### DIFF
--- a/app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
+++ b/app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
@@ -26,11 +26,13 @@ import com.example.taskyy.ui.events.RegisterEvent
 import com.example.taskyy.ui.events.ReminderEvent
 import com.example.taskyy.ui.screens.AgendaScreen
 import com.example.taskyy.ui.screens.EditTextScreen
+import com.example.taskyy.ui.screens.EventScreen
 import com.example.taskyy.ui.screens.LoginScreen
 import com.example.taskyy.ui.screens.RegisterScreen
 import com.example.taskyy.ui.screens.ReminderScreen
 import com.example.taskyy.ui.viewmodels.AgendaViewModel
 import com.example.taskyy.ui.viewmodels.EditTextViewModel
+import com.example.taskyy.ui.viewmodels.EventViewModel
 import com.example.taskyy.ui.viewmodels.LoginViewModel
 import com.example.taskyy.ui.viewmodels.RegisterViewModel
 import com.example.taskyy.ui.viewmodels.ReminderViewModel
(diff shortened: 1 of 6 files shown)


app/src/main/java/com/example/taskyy/domain/navigation/Navigation.kt
```diff
@@ -129,13 +131,34 @@ fun Nav() {
...
+                    navDeepLink {
+                        uriPattern =
+                            "tasky://www.myapp.com/eventScreen?zoneDateTime={zoneDateTime}" +
+                                    "&isEditingMode={isEditingMode}&agendaItemId={agendaItemId}"
+                    },
+                ),
+            ) {
+                val eventViewModel = hiltViewModel<EventViewModel>()
+                val state by eventViewModel.state.collectAsState()
+                val timeDateState by eventViewModel.timeAndDateState.collectAsState()
```
philipplackner, line 147: I'd recommend to use `collectAsStateWithLifecycle()` that since it's the safest way to collect flows in Android. For that, you'll need this dependency:
```
implementation "androidx.lifecycle:lifecycle-runtime-compose:2.x.x"
```
Then use it like this:
```
val state by viewModel.state.collectAsStateWithLifecycle()
```
If you have a simple flow it shouldn't make a difference, but as soon as you convert a normal `Flow` into a `StateFlow` with `stateIn`, the Flow operators will be executed even when the app is in the background (which you typically don't want when dealing with UI updates).

app/src/main/java/com/example/taskyy/ui/viewmodels/EventViewModel.kt
```diff
@@ -0,0 +1,84 @@
...
+
+}
+
+data class EventState(
+    val isAlarmSelectionExpanded: Boolean = false,
+    val isTimePickerSelectionExpanded: Boolean = false,
+    val isDatePickerExpanded: Boolean = false,
+    val alarmReminderTimeSelection: String = "",
+    val reminderType: ReminderType = ReminderType.ONE_HOUR_BEFORE,
+    val formattedReminderTime: Long = 3600000,
```
philipplackner, line 72: This value doesn't look formatted to me :D

app/src/main/java/com/example/taskyy/ui/viewmodels/EventViewModel.kt
```diff
@@ -0,0 +1,84 @@
...
+    val description: String = "Description",
+    val titleText: String = "New Event",
+    val saveFailedMessage: String = "",
+    val agendaItemType: AgendaItemType = AgendaItemType.EVENT_ITEM,
+    val eventId: String = "",
```
philipplackner, line 77: Instead of making non-existent IDs empty strings, I'd make them `null` instead. To me it's not obvious that just because an ID is empty there is no item. Don't be afraid to use nulls what they are intended to be used for (indicating that a value doesn't exist)

philipplackner (approved):

PR #38, opened 2024-04-19
made some updates and resolved more minor bugs. For some reason the g…

Mooky4591:
…etFullAgenda call gives me a "Unable to create call adapter" error. I don't know I've set up the response correctly

### DIFF
--- a/app/src/main/java/com/example/taskyy/data/remote/interceptors/ApiKeyInterceptor.kt
+++ b/app/src/main/java/com/example/taskyy/data/remote/interceptors/ApiKeyInterceptor.kt
@@ -1,10 +1,12 @@
-package com.example.taskyy.data.remote
+package com.example.taskyy.data.remote.interceptors
 
+import com.example.taskyy.domain.repository.UserPreferences
 import okhttp3.Interceptor
 import okhttp3.Response
 
 class ApiKeyInterceptor(
-    private val apiKey: String
+    private val apiKey: String,
+    private val userPreferences: UserPreferences
 ): Interceptor {
 
     override fun intercept(chain: Interceptor.Chain): Response {
@@ -13,6 +15,7 @@ class ApiKeyInterceptor(
                 .request()
                 .newBuilder()
                 .addHeader("x-api-key", apiKey)
+                .addHeader("token", userPreferences.getUserToken())
                 .build()
         )
     }
--- a/app/src/main/java/com/example/taskyy/data/remote/interceptors/TokenInterceptor.kt
+++ b/app/src/main/java/com/example/taskyy/data/remote/interceptors/TokenInterceptor.kt
@@ -1,0 +1,19 @@
+package com.example.taskyy.data.remote.interceptors
+
+import com.example.taskyy.domain.repository.UserPreferences
+import okhttp3.Interceptor
+import okhttp3.Response
+
+class TokenInterceptor(
+    private val userPreferences: UserPreferences
+) : Interceptor {
+    override fun intercept(chain: Interceptor.Chain): Response {
+        return chain.proceed(
+            chain
+                .request()
+                .newBuilder()
+                .addHeader("token", userPreferences.getUserToken())
+                .build()
+        )
+    }
+}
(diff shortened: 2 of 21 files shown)


app/src/main/java/com/example/taskyy/data/remote/response_objects/EventResponse.kt
```diff
@@ -0,0 +1,10 @@
+package com.example.taskyy.data.remote.response_objects
+
+data class EventResponse(
+    val id: String,
+    val title: String,
+    val description: String,
+    val time: Long,
+    val remindAt: Long,
+    val isDone: Boolean
+)
```
philipplackner, line 10: This does not represent an event, make sure to stick to the docs and adjust this, then you also won't get a call adapter error :)

app/src/main/java/com/example/taskyy/ui/screens/LoginScreen.kt
```diff
@@ -58,17 +58,42 @@ fun LoginScreen(state: LoginState, onEvent: (LoginEvent) -> Unit) {
...
+                            loginObject?.let { it1 ->
+                                LoginEvent.OnLoginClick(
+                                    it1
+                                )
+                            }?.let { it2 -> onEvent(it2) }
+                        }, text = stringResource(
+                            id = R.string.login
+                        ), login = loginObject
+                    )
+                }
```
philipplackner: Having to construct this login object here seems very unnecessary and overly complex to me given that you have all these `it`, `it1` and `it2` parameters which don't say anything. Why not just pass the email and pw to the VM?

app/src/main/java/com/example/taskyy/ui/viewmodels/LoginViewModel.kt
```diff
@@ -90,14 +90,14 @@ class LoginViewModel @Inject constructor(
...
-    var password: String = "",
-    var name: String = "",
-    var isLogginIn: Boolean = false,
-    var isEmailValid: Boolean = false,
-    var isPasswordVisible: Boolean = false,
-    var isLoginSuccessful: Boolean = false,
-    var loginErrorMessage: UiText? = null,
-    var isReady: Boolean = false
+    val email: String? = null,
+    val password: String? = null,
```
philipplackner: When text field values are empty it's very clear that it refers to an empty text field, since it's a valid state for a text field value (unlike an empty string for an error message which is not a valid error). So here, the default values can stay empty strings. Null values would rather give me the feel that the whole text field wouldn't exist in that case

philipplackner (approved):

app/src/main/java/com/example/taskyy/ui/screens/LoginScreen.kt
Mooky4591: That was what I was doing originally but I tried to make my strings nullable in the view model state but then it required all this .let extra stuff to resolve the errors. Then when I tried to get rid of them and use the !! operator instead, the layout of my login screen got messed up.

app/src/main/java/com/example/taskyy/data/remote/interceptors/ApiKeyInterceptor.kt
```diff
@@ -13,6 +15,7 @@ class ApiKeyInterceptor(
                 .request()
                 .newBuilder()
                 .addHeader("x-api-key", apiKey)
+                .addHeader("token", userPreferences.getUserToken())
```
philipplackner, line 18: If this interceptor apparently does more than just adding the API key, I'd also communicate that with its name. You could just call it `AuthInterceptor` instead

app/src/main/java/com/example/taskyy/data/remote/interceptors/TokenInterceptor.kt
```diff
@@ -0,0 +1,19 @@
...
+    override fun intercept(chain: Interceptor.Chain): Response {
+        return chain.proceed(
+            chain
+                .request()
+                .newBuilder()
+                .addHeader("token", userPreferences.getUserToken())
+                .build()
+        )
+    }
+}
```
philipplackner, line 19: And then this one seems unnecessary

philipplackner (approved):

